PR #4, opened 2024-05-16
Add different manifests and instructions for dev install

daurer:
- [x] Add dev-values.yaml for development environments using less resources and replicas
- [x] Update README
- [x] Clean up and test dev-values.yaml for workflows
- [x] Clean up and dev-values.yaml for vcluster 
- [x] Bumped version of vcluster and workflows

### DIFF
--- a/charts/workflows-cluster/dev-values.yaml
+++ b/charts/workflows-cluster/dev-values.yaml
@@ -1,0 +1,29 @@
+vcluster:
+  controlPlane:
+    coredns:
+      deployment:
+        replicas: 1
+        resources:
+          limits:
+            cpu: 200m
+            memory: 500Mi
+          requests:
+            cpu: 100m
+            memory: 250Mi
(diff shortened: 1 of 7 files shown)


charts/workflows-cluster/dev-values.yaml
garryod, line 1: This can be the diff between the desired dev values and the `values.yaml` as this will override

charts/workflows/dev-values.yaml
```diff
@@ -0,0 +1,54 @@
...
+    resources:
+      limits:
+        cpu: 500m
+        memory: 512Mi
+      requests:
+        cpu: 250m
+        memory: 256Mi
+
+postgresql-ha:
+  enabled: true
```
garryod, line 34: Nit: Can remove this line

garryod (approved):

PR #8, opened 2024-05-17
Add event controllers

daurer:
- [x] Argo event controllers
- [x] EventBus deployments
- [x] EventSource deployments

The EventBus deployments are not deployed with the argo-events helm chart: https://github.com/argoproj/argo-helm/issues/840

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,11 @@ Secondly, deploy the workflows service in the virtual cluster:
 vcluster connect workflows-cluster -- helm install workflows charts/workflows -n workflows --create-namespace
 ```
 
+Finally, deploy the events service in the virtual cluster:
+```sh
+vcluster connect workflows-cluster -- helm install workflows charts/events -n events --create-namespace
+```
+
 ## Deployment in developer mode
 
 First, deploy the workflows virtual cluster using the developer manifest:
@@ -23,11 +28,16 @@ First, deploy the workflows virtual cluster using the developer manifest:
 helm install workflows-cluster charts/workflows-cluster -f charts/workflows-cluster/dev-values.yaml
 ```
 
-Secondly, deploy the workflows service in the virtual cluster using the developer manifest :
+Secondly, deploy the workflows service in the virtual cluster using the developer manifest:
 ```sh
 vcluster connect workflows-cluster -- helm install workflows charts/workflows -n workflows -f charts/workflows/dev-values.yaml --create-namespace
 ```
 Note that for getting the workflows-server to run inside the dev environment it is necessary to extract the argo-server-sso secret, delete the deployed sealed secret and then deploy a new sealed secret using ```kubectl create -f <SEALED-SECRET>``` inside the virtual cluster.
+
+Finally, deploy the events service in the virtual cluster using the developer manifest:
+```sh
+vcluster connect workflows-cluster -- helm install workflows charts/events -n events -f charts/events/dev-values.yaml --create-namespace
+```
 
 ## Serve Docs
 
--- a/charts/events/templates/event-bus.yaml
+++ b/charts/events/templates/event-bus.yaml
@@ -1,0 +1,28 @@
+# https://raw.githubusercontent.com/argoproj/argo-events/stable/examples/eventbus/native.yaml
+{{- if .Values.argoevents.enabled }}
+  {{- range $eventBus := $.Values.eventBuses }}
+---
+apiVersion: argoproj.io/v1alpha1
+kind: EventBus
+metadata:
+  namespace: {{ $eventBus.namespace }}
+  name: {{ $eventBus.name }}
+spec:
+  nats:
+    native:
+      # Optional, defaults to 3. If it is < 3, set it to 3, that is the minimal requirement.
+      replicas: 3
+      # Optional, auth strategy, "none" or "token", defaults to "none"
+      auth: token
+      {{- if $eventBus.resources }}  
+      containerTemplate:
+        resources:
+          {{- $eventBus.resources | toYaml | nindent 10 }} 
+      {{- end }}
+      {{- if $eventBus.metricsresources }}  
+      metricsContainerTemplate:
+        resources:
+           {{- $eventBus.metricsresources | toYaml | nindent 10 }} 
+      {{- end }}
+  {{- end }}
+{{- end }}
--- a/charts/events/templates/event-sources.yaml
+++ b/charts/events/templates/event-sources.yaml
@@ -1,0 +1,18 @@
+{{- if .Values.argoevents.enabled }}
+  {{- range $eventSource := $.Values.eventSources }}
+---
+apiVersion: argoproj.io/v1alpha1
+kind: EventSource
+metadata:
+  name: {{ $eventSource.name }}
+spec:
+  service:
+    ports:
+      - port: {{$eventSource.port }}
+        targetPort: {{$eventSource.port }}
+  {{- if $eventSource.webhook }} 
+  webhook:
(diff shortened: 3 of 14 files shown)


README.md
```diff
@@ -40,4 +50,3 @@ pipx runpip mkdocs install -r docs/requirements.txt
 Now, serve the docs with `mkdocs`:
 ```sh
 mkdocs serve
-```
```
garryod, line 43: This seems to have been deleted in error

charts/events/charts/argo-events-2.4.4.tgz
garryod, line 1: This needs to be `.gitignore`d

charts/events/templates/event-bus.yaml
```diff
@@ -0,0 +1,36 @@
...
+  namespace: {{ $eventBus.namespace }}
+  name: {{ $eventBus.name }}
+spec:
+  nats:
+    native:
+      # Optional, defaults to 3. If it is < 3, set it to 3, that is the minimal requirement.
+      replicas: 3
+      # Optional, auth strategy, "none" or "token", defaults to "none"
+      auth: token
+      {{- if eq ($eventBus.environment | default "prod") "dev" }}
```
garryod, line 17: I would prefer if we took the individual values from the `values.yaml` and over-rid them with `dev-values.yaml`. This bakes the idea of `values` and `dev-values` into the chart, where they don't belong

charts/events/templates/sensor-service-account.yaml
garryod, line 1: Do we need multiple of these in separate namespaces? One for each science group

charts/events/templates/workflow-role-binding.yaml
garryod, line 1: Do we need multiple of these in separate namespaces? One for each science group

charts/events/values.yaml
```diff
@@ -0,0 +1,31 @@
...
+  enabled: true
+  controller:
+    replicas: 1
+    pdb:
+      minAvailable: 1
+    serviceAccount:
+      name: argo-events-sa
+    # securityContext:
+    #   runAsNonRoot: true
+    #   runAsUser: 9731
```
garryod, line 11: Should this have been included?

docs/index.md
```diff
@@ -1,3 +1,46 @@
...
+
+```sh
+vcluster connect workflows-cluster -- kubectl -n events get workflow
+```
+
+We can also use the ```argo``` CLI to look at the logs of the latest workflow
+
+```sh
+vcluster connect workflows-cluster -- argo logs -n events @latest
+```
```
garryod, line 46: Can we move this to a separate file? E.g. `./how-tos/create-trigger` or `./tutorials/trigger-with-webhook`?

garryod (changes requested):

charts/events/values.yaml
daurer: not needed, removed it

charts/events/templates/event-sources.yaml
```diff
@@ -0,0 +1,25 @@
+{{- if index .Values "argo-events" "enabled" }}
+  {{- range $eventSource := $.Values.eventSources }}
+    {{- if eq $eventSource.name "webhook" }} 
+---
+apiVersion: argoproj.io/v1alpha1
+kind: EventSource
+metadata:
+  name: webhook
```
garryod, line 8: This has to be unique / templated

charts/events/templates/event-sources.yaml
```diff
@@ -0,0 +1,25 @@
...
+        targetPort: {{$eventSource.port }}
+  webhook:
+    # event-source can run multiple HTTP servers. Simply define a unique port to start a new HTTP server
+    example:
+      # port to run HTTP server on
+      port: "{{$eventSource.port }}"
+      # endpoint to listen to
+      endpoint: /{{ $eventSource.endpoint }}
+      # HTTP request method to allow. In this case, only POST requests are accepted
+      method: POST
```
garryod, line 22: Perhaps something like this?
```suggestion
  {{- if $eventSource.webhook }}
  webhook:
    {{- $eventSource.webhook | toYaml | nindent 6 }}
  {{- end }}
```

docs/how-tos/trigger-with-webhooks.md
```diff
@@ -0,0 +1,38 @@
+# How to trigger a workflow using a webhook sensor
```
garryod, line 1: This needs to fit in the side-nav, so best kept short, pehaps this?
```suggestion
# Trigger with a Webhook Sensor
```

garryod (changes requested):

charts/events/templates/event-sources.yaml
```diff
@@ -0,0 +1,18 @@
+{{- if index .Values "argo-events" "enabled" }}
```
garryod, line 1: Best just to traverse the object
```suggestion
{{- if .Values.argo-events.enabled }}
```

docs/how-tos/trigger-with-webhooks.md
garryod, line 1: Would be good to use the proper domains (e.g. `workflows.diamond.ac.uk/triggering/*`) instead of vcluster access - though we might need to set up an ingress before we can do so

garryod (changes requested):

charts/events/templates/sensor-service-account.yaml
daurer: Perhaps, let's discuss in the next sprint planning meeting. Would say out of scope for this PR.

charts/events/templates/event-sources.yaml
daurer: yeah, this would be cleaner but helm is not happy with `bad character U+002D '-' `
See also here: https://stackoverflow.com/questions/63853679/helm-templating-doesnt-let-me-use-dash-in-names

charts/events/values.yaml
```diff
@@ -0,0 +1,36 @@
+argo-events:
+  enabled: true
+  controller:
+    replicas: 1
```
garryod, line 4: This should probably be greater than one, a PDB of 1 doesn't really work with one replica

mkdocs.yml
```diff
@@ -1,4 +1,6 @@
 site_name: Workflows
 plugins:
   - techdocs-diamond
+markdown_extensions:
+  - admonition
```
garryod, line 5: This is included in `techdocs-diamond`

garryod (changes requested):

mkdocs.yml
daurer: cool, thanks

charts/events/values.yaml
daurer: increased to 3

daurer: Actually, events with replicas > 1 seems to cause "Failed to load global configuration file" on the controller pod. Reverted back to single replica for now. We can look into this again as part of HA for events triggering

garryod: Sounds reasonable, can you stick an issue in to make events HA?

charts/events/templates/event-bus.yaml
```diff
@@ -0,0 +1,28 @@
+# https://raw.githubusercontent.com/argoproj/argo-events/stable/examples/eventbus/native.yaml
+{{- if index .Values "argo-events" "enabled" }}
```
garryod, line 2: Would be good to remove all the `index` based stuff

charts/events/values.yaml
daurer: see #14

garryod (approved):
LGTM

PR #11, opened 2024-05-23
Add Sessionspaces namespace-controller for vcluster

iamvigneshwars:
- [x] create/delete session namespace
- [x] batch query LDAP database to get gid
- [x] create configmap with gid and session info
- [x] helm chart


.devcontainer/Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+    KREW="krew-${OS}_${ARCH}" && \
+    curl -fsSLO "https://github.com/kubernetes-sigs/krew/releases/latest/download/${KREW}.tar.gz" && \
+    tar zxvf "${KREW}.tar.gz" && \
+    ./"${KREW}" install krew && \
+    mv "${KREW}" /usr/local/bin/krew
+
+# Add krew to the PATH
+ENV PATH="${KREW_ROOT:-$HOME/.krew}/bin:$PATH"
+
+RUN krew install oidc-login
```
garryod, line 23: Why?

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+
+# Download and install vcluster
+RUN wget https://github.com/loft-sh/vcluster/releases/download/v0.19.5/vcluster-linux-amd64 -O /usr/local/bin/vcluster && \
+    chmod +x /usr/local/bin/vcluster
+
+RUN mv /usr/local/bin/vcluster /bin/vcluster
```
garryod, line 29:
```suggestion
RUN wget https://github.com/loft-sh/vcluster/releases/download/v0.19.5/vcluster-linux-amd64 -O /bin/vcluster && \
    chmod +x /usr/local/bin/vcluster
```

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+
+COPY ./config_argus /root/.kube/config
+
+COPY ./ca.crt /root/ca.crt
```
garryod, line 33: Ideally we wouldn't copy these in from a local path... Any better alternatives? Could we mount them in from a known path? At the very least we could provide instructions for getting them in the `README.md`

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,25 @@
...
+      "extensions": [
+        "rust-lang.rust-analyzer",
+        "tamasfe.even-better-toml",
+        "ms-kubernetes-tools.vscode-kubernetes-tools"
+      ]
+    }
+  },
+  // "features": {
+  //   "ghcr.io/devcontainers/features/common-utils:2.4.2": {}
+  // },
```
garryod, line 18: Can remove this if not using

.devcontainer/devcontainer-lock.json
garryod, line 1: This doesn't agree with the `devcontainer.json`

.github/workflows/container.yaml
garryod, line 1: Can we rename to `sessionspaces-container.yaml`? Feel like we may eventually have multiple containers for things

.gitignore
```diff
@@ -1,2 +1,9 @@
+.env
+# k8s configs
+ca.crt
+config_argus
+# vcluster config
+kubeconfig.yaml
+/target
```
garryod, line 7:
```suggestion
# Rust
/target
```

Cargo.toml
garryod, line 1: Do we want a workspace yet?

Dockerfile
garryod, line 1: Can we move this into `sessionspaces/`?

charts/sessionspaces/Chart.yaml
garryod, line 1: Do we need / want to nest these charts?

charts/sessionspaces/charts/sessionspaces/templates/clusterrole.yaml
```diff
@@ -0,0 +1,42 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: sessionspaces-clusterrole
```
garryod, line 4: We don't need to affix the kind here. Would probably be best just to use the `sessionspaces.fullname`

charts/sessionspaces/charts/sessionspaces/templates/clusterrolebinding.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: sessionspaces-clusterrolebinding
```
garryod, line 4: We don't need to affix the kind here. Would probably be best just to use the `sessionspaces.fullname`

charts/sessionspaces/charts/sessionspaces/templates/clusterrole.yaml
```diff
@@ -0,0 +1,42 @@
...
+    "workflowtasksets/status",
+    "clusterworkflowtemplates",
+    "cronworkflows",
+    "eventsources",
+    "sensors",
+    "workfloweventbindings",
+    "workflows",
+    "workflowtemplates"
+  ]
+  verbs: ["get", "list", "watch", "patch", "create"]
```
garryod, line 42: Not sure I understand why these are the rules for the `sessionspaces` ServiceAccount, surely it only needs to be able to `create` and `delete`; `namespace`s, `role`s, `serviceaccount`s, and `rolebinding`s?

charts/sessionspaces/charts/sessionspaces/templates/service.yaml
garryod, line 1: Why do we need a service?

charts/sessionspaces/charts/sessionspaces/templates/deployment.yaml
```diff
@@ -0,0 +1,59 @@
...
+              valueFrom:
+                secretKeyRef:
+                  name: {{ .Values.database.password.secretName }}
+                  key: {{ .Values.database.password.secretKey }}
+            - name: DATABASE_URL
+              value: {{ include "sessionspaces.databaseURL" . }}
+          ports:
+            - name: http
+              containerPort: {{ .Values.service.port }}
+              protocol: TCP
```
garryod, line 45: What does this port expose?

charts/sessionspaces/charts/sessionspaces/values.yaml
```diff
@@ -0,0 +1,47 @@
...
+database:
+  host: ""
+  user: ""
+  password:
+    secretName: ""
+    secretKey: ""
+
+opa:
+  url: ""
+
```
garryod, line 23: This can be removed

charts/sessionspaces/charts/sessionspaces/values.yaml
```diff
@@ -0,0 +1,47 @@
...
+replicaCount: 1
+
```
garryod, line 25: Do we support replication? Isn't the controller stateful? Should we aim for HA? If so, how?

sessionspaces/README.md
```diff
@@ -0,0 +1,4 @@
+# SessionSpaces
+
+A service which templates out configuration (particularly namespaces)
+for each Session in ISPyB
```
garryod, line 4: Would be good to have a little bit more detail as to what this does. I.e. what k8s resources are created, and the ISPyB tables that are referenced in doing so

sessionspaces/src/main.rs
```diff
@@ -0,0 +1,162 @@
...
+            if let Some(members) = sessions.get(namespace) {
+                create_visit_member_service_account(
+                    namespace.clone(),
+                    members.clone(),
+                    k8s_client.clone(),
+                )
+                .await?;
+            }
+        } else if members.is_some() {
+            if current_members.is_none() || current_members.as_ref() != Some(&members.unwrap()) {
```
garryod, line 144: Could use `is_some_and` here to avoid the `unwrap`

sessionspaces/src/permissionables/session.rs
```diff
@@ -0,0 +1,77 @@
...
+    /// The proposal number with which this session is associated
+    pub proposal: u32,
+    /// The visit number within the proposal
+    pub visit: u32,
+}
+
+impl Session {
+    /// Fetches [`Session`]s from ISPyB
+    #[instrument(name = "fetch_sessions")]
+    pub async fn fetch(ispyb_pool: &MySqlPool) -> Result<Vec<Self>, sqlx::Error> {
```
garryod, line 19: This needs to be tested. Can do it with a mock database instance as I did [in the authz bundler](https://github.com/DiamondLightSource/authz/blob/6539c4d95bb83724bfdf4ea1b54f7040efe5e191/bundler/src/permissionables/sessions.rs#L103)

sessionspaces/src/resources/member_sa.rs
```diff
@@ -0,0 +1,135 @@
...
+};
+use std::collections::BTreeMap;
+use tracing::instrument;
+
+/// The Resource name for the visit member ServiceAccount, Role & RoleBinding
+const VISIT_MEMBER: &str = "visit-member";
+
+/// Creates or updates a ClusterRole for use by the argo-workflows ServiceAccounts
+#[instrument(skip(k8s_client))]
+pub async fn create_visit_member_role(k8s_client: kube::Client) -> Result<(), kube::Error> {
```
garryod, line 19: Need to figure out a way of testing these

garryod (changes requested):
Looking pretty good, the development environment and deployment have room to be simplified in a few places, and definitely needs some tests, but the actual code looks good to go

.devcontainer/Dockerfile
iamvigneshwars: Can't use kubectl inside dev container without it.

charts/sessionspaces/charts/sessionspaces/templates/clusterrole.yaml
iamvigneshwars: Yeah I started with those simple rules, but I ran into permission errors, I've added most of the rules from those errors, But yeah, I think we can remove some of the unwanted rules

charts/sessionspaces/charts/sessionspaces/templates/service.yaml
iamvigneshwars: Ahh, I templated it from our graph helm charts,.. will remove all the unwanted stuff!

sessionspaces/src/main.rs
iamvigneshwars: makes sense!

.devcontainer/Dockerfile
iamvigneshwars: I'm not sure if Docker would allow to mount it from a global path, I tried it and it didn't work! 
But yeah we could provide instructions for getting them.

Cargo.toml
iamvigneshwars: We might have any other service in the future? I thought we could just use a standard dev workflow!

charts/sessionspaces/templates/clusterrole.yaml
```diff
@@ -0,0 +1,39 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: {{ include "sessionspaces.fullname" . }}
+rules:
+- apiGroups: [""]
+  resources: ["pods", "services"]
+  verbs: ["get", "list", "watch", "patch", "create"]
```
garryod, line 8: Does sessionspaces need permission to operate on `pods` or `services`?

charts/sessionspaces/templates/clusterrole.yaml
```diff
@@ -0,0 +1,39 @@
...
+    "workflowtasksets/status",
+    "clusterworkflowtemplates",
+    "cronworkflows",
+    "eventsources",
+    "sensors",
+    "workfloweventbindings",
+    "workflows",
+    "workflowtemplates"
+  ]
+  verbs: ["get", "list", "watch", "patch", "create"]
```
garryod, line 39: Sessionspaces shouldn't do anything with argo workflows resources

charts/sessionspaces/templates/clusterrole.yaml
```diff
@@ -0,0 +1,39 @@
...
+  verbs: ["get", "list", "watch", "patch", "create"] 
+- apiGroups: [""]
+  resources: ["namespaces"]
+  verbs: ["create", "delete", "get", "list"]
+- apiGroups: [""]
+  resources: ["pods/exec"]
+  verbs: ["create"]
+- apiGroups: [""]
+  resources: ["pods/logs"]
+  verbs: ["get", "watch"]
```
garryod, line 17: Does sessionspaces need permission to exec into pods (`pods/exec`) or read pods logs (`pods/logs`)?

charts/sessionspaces/templates/clusterrole.yaml
garryod, line 1: This should only be created if we've set `serviceAccount.create`

charts/sessionspaces/templates/clusterrolebinding.yaml
garryod, line 1: This should only be created if we've set `serviceAccount.create`

charts/sessionspaces/templates/clusterrole.yaml
```diff
@@ -0,0 +1,39 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: {{ include "sessionspaces.fullname" . }}
```
garryod, line 4: Maybe we should use the `sessionspaces.serviceAccountName` here? Thoughts?

sessionspaces/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+
+ARG DATABASE_URL
+
+WORKDIR /app
+
+COPY . .
+
+RUN cargo build --release
+
+FROM gcr.io/distroless/cc AS deploy
```
garryod, line 11: Want a tag (hash) here

sessionspaces/Dockerfile
garryod, line 1: Probably need a `.dockerignore` so we're not copying `/target` etc.

garryod (changes requested):

charts/sessionspaces/templates/clusterrole.yaml
iamvigneshwars: Done!

charts/sessionspaces/templates/clusterrole.yaml
iamvigneshwars: Yeah, makes sense!

charts/sessionspaces/templates/clusterrole.yaml
iamvigneshwars: It creates a argo-workflows clusterrole and has rules for pods, but not for services, I will update it.

charts/sessionspaces/templates/clusterrole.yaml
iamvigneshwars: It creates a argo-workflows ClusterRole with those resources

charts/sessionspaces/templates/clusterrole.yaml
iamvigneshwars: It creates two clusterroles (visit-member, argo-workflows) with these resources.

charts/sessionspaces/templates/clusterrolebinding.yaml
iamvigneshwars: Done!

sessionspaces/src/resources/workflows_sa.rs
garryod, line 1: Can remove this in favour of creating the `workflows` & `triggering` specific resources with Kyverno

sessionspaces/src/resources/member_sa.rs
garryod, line 1: Can remove this in favour of creating the `workflows` & `triggering` specific resources with Kyverno

sessionspaces/src/resources/config_maps.rs
```diff
@@ -0,0 +1,93 @@
...
+use ldap3::{LdapConnAsync, Scope, SearchEntry};
+use std::{
+    collections::{BTreeMap, BTreeSet},
+    error::Error,
+};
+use tracing::{info, instrument};
+
+const POLICY_CONFIG: &str = "policy-config";
+
+pub async fn ldap_search(namespace: String) -> Result<String, Box<dyn Error>> {
```
garryod, line 15: This should probably be part of the `permissionables` module as it's getting data

garryod (changes requested):

charts/sessionspaces/templates/clusterrole-argo-workflows.yaml
garryod, line 1: We can move this to `charts/workflows/templates` - may be best in a different PR

charts/sessionspaces/templates/clusterrole-visit-member.yaml
garryod, line 1: We can move this to `charts/workflows/templates` - may be best in a different PR

charts/sessionspaces/templates/clusterrolebinding.yaml
```diff
@@ -0,0 +1,15 @@
...
+metadata:
+  name: {{ include "sessionspaces.fullname" . }}
+  namespace: kube-system
+subjects:
+- kind: ServiceAccount
+  name: {{ include "sessionspaces.serviceAccountName" . }}
+  namespace: kube-system
+roleRef:
+  kind: ClusterRole
+  name: {{ include "sessionspaces.fullname" . }}
```
garryod, line 13: This needs to match `metadata.name` in the associated `ClusterRole`
```suggestion
  name: {{ include "sessionspaces.serviceAccountName" . }}
```

charts/sessionspaces/templates/clusterrolebinding.yaml
```diff
@@ -0,0 +1,15 @@
+{{- if .Values.serviceAccount.create -}}
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: {{ include "sessionspaces.fullname" . }}
```
garryod, line 5:
```suggestion
  name: {{ include "sessionspaces.serviceAccountName" . }}
```

sessionspaces/Dockerfile
garryod, line 1: Is it worth doing the `copy Cargo.toml Cargo.lock` trick to facilitate caching of container builds?

sessionspaces/src/permissionables/ldap.rs
```diff
@@ -0,0 +1,27 @@
+use ldap3::{LdapConnAsync, Scope, SearchEntry};
+
+pub async fn ldap_search(namespace: String) -> Option<String> {
```
garryod, line 3: Probably want to bring this inside `Session::fetch`

garryod (changes requested):

sessionspaces/src/permissionables/ldap.rs
```diff
@@ -0,0 +1,42 @@
+use crate::SessionSpaces;
+use ldap3::{LdapConnAsync, Scope, SearchEntry};
+
+pub async fn update_gid(sessions: &mut SessionSpaces) {
+    let (conn, mut ldap) = LdapConnAsync::new("ldap://ldap.diamond.ac.uk")
```
garryod, line 5: Definitely want to pass this in from the command line

sessionspaces/src/permissionables/ldap.rs
garryod, line 1: Probably want to name this something related to what it fetches - perhaps `posix_attributes.rs`?

sessionspaces/src/permissionables/ldap.rs
```diff
@@ -0,0 +1,42 @@
...
+        .await
+        .unwrap();
+    ldap3::drive!(conn);
+    let session_names = sessions.keys().cloned().collect::<Vec<_>>();
+    for session_chunk in session_names.chunks(1000) {
+        let mut filter = String::from("(&(objectclass=posixgroup)(|");
+        for cn in session_chunk {
+            filter.push_str(&format!("(cn={})", cn.replace("-", "_")));
+        }
+        filter.push_str("))");
```
garryod, line 15: Do we need to construct this query or can we just remove the `cn=` and get all the posix groups in `Groups.diamond.ac.uk`?

sessionspaces/src/permissionables/ldap.rs
```diff
@@ -0,0 +1,42 @@
...
+            .search(
+                "ou=Group,dc=diamond,dc=ac,dc=uk",
+                Scope::Subtree,
+                &filter,
+                vec!["gidnumber", "cn"],
+            )
+            .await
+            .unwrap()
+            .success()
+            .unwrap();
```
garryod, line 26: Are these `unwrap`s safe?

sessionspaces/src/permissionables/ldap.rs
```diff
@@ -0,0 +1,42 @@
...
+        for entry in rs {
+            if let Some(gid) = SearchEntry::construct(entry.clone()).attrs.get("gidNumber") {
+                let session_name = SearchEntry::construct(entry)
+                    .attrs
+                    .get("cn")
+                    .unwrap()
```
garryod, line 32: Is this `unwrap` safe?

sessionspaces/src/permissionables/ldap.rs
```diff
@@ -0,0 +1,42 @@
+use crate::SessionSpaces;
+use ldap3::{LdapConnAsync, Scope, SearchEntry};
+
+pub async fn update_gid(sessions: &mut SessionSpaces) {
```
garryod, line 4: It would likely be neater to have a method which fetched all the posix attributes of groups - like `fetch` in the other `permissionables`  - and combine these downstream. Means that we can later drop the ISPyB ones when move stuff is in ldap

sessionspaces/src/resources/config_maps.rs
```diff
@@ -0,0 +1,64 @@
...
+            );
+        }
+        None => {}
+    };
+    match gid_number {
+        Some(gid) => {
+            configmap_data.insert("gid".to_string(), gid);
+        }
+        None => {}
+    }
```
garryod, line 43: Would be neater as an `if let Some(gid) = gid_number { ... }`

sessionspaces/src/resources/config_maps.rs
```diff
@@ -0,0 +1,64 @@
...
+        Some(members_set) => {
+            let quoted_members: Vec<String> =
+                members_set.iter().map(|s| format!("\"{}\"", s)).collect();
+            configmap_data.insert(
+                "members".to_string(),
+                format!("[{}]", quoted_members.join(", ")),
+            );
+        }
+        None => {}
+    };
```
garryod, line 37: Would be neater as an `if let Some(members) = members_set { ... }`

sessionspaces/src/resources/config_maps.rs
```diff
@@ -0,0 +1,64 @@
+use k8s_openapi::api::core::v1::ConfigMap;
+use kube::{
+    api::{ObjectMeta, Patch, PatchParams},
+    Api, Client,
+};
+use std::collections::{BTreeMap, BTreeSet};
+use tracing::{info, instrument};
+
+const POLICY_CONFIG: &str = "policy-config";
```
garryod, line 9: Would be good to have it named something Sessionspace related, maybe even just `sessionspace`

garryod (changes requested):

sessionspaces/src/permissionables/ldap.rs
iamvigneshwars: and keep this url as default?

sessionspaces/src/permissionables/ldap.rs
iamvigneshwars: We might query many unwanted namespaces?

sessionspaces/src/permissionables/ldap.rs
iamvigneshwars: It is safe if the query is built properly

sessionspaces/src/permissionables/ldap.rs
iamvigneshwars: makes sense

sessionspaces/src/permissionables/ldap.rs
iamvigneshwars: But yeah, if we want to separate ldap and sessions then fetching the everything from ldap makes sense!

sessionspaces/src/main.rs
```diff
@@ -0,0 +1,214 @@
...
+        let members = match session_info {
+            Some(sess_info) => {
+                if !sess_info.members.is_empty() {
+                    Some(sess_info.members.clone())
+                } else {
+                    None
+                }
+            }
+            None => None,
+        };
```
garryod, line 163:
```suggestion
        let members = session_info.map(|info| info.members).flatten();
```

sessionspaces/src/main.rs
```diff
@@ -0,0 +1,214 @@
...
+        let current_members = if let Some(current_session_info) = current_sesssion_info {
+            if !current_session_info.members.is_empty() {
+                Some(current_session_info.members.clone())
+            } else {
+                None
+            }
+        } else {
+            None
+        };
```
garryod, line 172:
```suggestion
        let current_members = current_session_info.map(|info| info.members).flatten();  
```

garryod (dismissed):

garryod (approved):

PR #16, opened 2024-05-30
Move argo workflows server ingress to workflows chart

garryod:
Just cleaning things up, was originally in the delayed #13, but is useful now


daurer (approved):
Looks good to me

PR #22, opened 2024-06-12
Add `managed-by` label to sessionspaces resources

garryod:
Adds the `app.kubernetes.io/managed-by` label to resources created by `sessionspaces`. Any thoughts on whether we should add any of the other [recommended labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/)?


iamvigneshwars:
I think we only need the managed by label right now!

iamvigneshwars (approved):

PR #24, opened 2024-06-13
Add proposal and `ixx_admin` admin members to sessionspace membership

garryod:


sessionspaces/tests/migrations/0_createISPyBTables.sql
iamvigneshwars, line 1: EOF

iamvigneshwars (approved):

PR #27, opened 2024-06-17
add pod securityContext clusterpolicy

iamvigneshwars:


charts/workflows/templates/pod-securitycontext-clusterpolicy.yaml
```diff
@@ -0,0 +1,24 @@
...
+      context:
+        - name: values
+          configMap:
+            name: sessionspaces
+            namespace: "{{ `{{request.object.metadata.namespace}}` }}"
+      mutate:
+        patchStrategicMerge:
+          spec:
+            securityContext:
+              runAsGroup: "{{ `{{values.data.gid | parse_json(@).to_number(@)}}` }}"
```
garryod, line 24: I think we also want to set the `runAsUser` to a value set in the `values.yaml`. We may also want to reject any pods with `allowPrivilegeEscalation` set to `true`

charts/workflows/templates/pod-securitycontext-clusterpolicy.yaml
garryod, line 1: Do we need to set `spec.containers[].securityContext` too? Or is it covered by `spec.securityContext`?

garryod (changes requested):

charts/workflows/templates/pod-securitycontext-clusterpolicy.yaml
iamvigneshwars: `spec.securityContext` covers it

charts/workflows/test-policy/pod-securitycontext/chainsaw-test.yaml
```diff
@@ -0,0 +1,46 @@
...
+              image: nginx:latest
+    - assert:
+        resource:
+          apiVersion: v1
+          kind: Pod
+          metadata:
+            namespace: test
+          spec:
+            securityContext:
+              runAsGroup: 1234
```
garryod, line 46: We also want to test the `runAsUser` is set appropriately and that setting `allowPrivilegeEscalation` results in a rejection

garryod (changes requested):

garryod (approved):

PR #29, opened 2024-06-19
Add hostpath to sessionspaces ConfigMap

iamvigneshwars:
If the sessions has a hostPath of structure `/dls/beamline/data/year/session` then it is added as `hostPath` in the sessionspaces ConfigMap


sessionspaces/src/resources/config_maps.rs
```diff
@@ -25,15 +25,26 @@ pub async fn create_configmap(
...
             "members".to_string(),
             serde_json::to_string(&session.members)?,
         ),
+        ("start_date".to_string(), session.start_date.to_string()),
+        ("end_date".to_string(), session.end_date.to_string()),
     ]);
     if let Some(gid) = session.gid {
         configmap_data.insert("gid".to_string(), gid);
     }
+    if BL_WITH_PATH.contains(&&session.beamline[..]) {
```
garryod, line 39: Just had a play... this works and is (in my opinion) more readable:
```suggestion
    if BL_WITH_PATH.contains(&session.beamline.as_str()) {
```

garryod (changes requested):

garryod (approved):

PR #30, opened 2024-06-19
add hostpath clusterpolicy

iamvigneshwars:


charts/workflows/templates/hostpath-clusterpolicy.yaml
```diff
@@ -0,0 +1,27 @@
...
+          namespaceSelector:
+            matchLabels:
+              app.kubernetes.io/managed-by: sessionspaces
+      context:
+        - name: values
+          configMap:
+            name: sessionspaces
+            namespace: "{{ `{{request.object.metadata.namespace}}` }}"
+      validate:
+        message: This hostPath is not allowed
```
garryod, line 22: Could be more explicit about what the path should be - bit better for someone who experiences a denial
```suggestion
        message: The hostPath must point to the session directory
```

charts/workflows/test-policy/hostpath/chainsaw-test.yaml
garryod, line 1: Perhaps this should be `pod-hostpath` to keep inline with the other test names?

charts/workflows/templates/hostpath-clusterpolicy.yaml
garryod, line 1: Can we use Kyverno [auto-gen rules](https://kyverno.io/docs/writing-policies/autogen/) to apply this to Argo `Workflow` resources? Would move the error message forward which would be nice

garryod (changes requested):

garryod (approved):
LGTM

PR #44, opened 2024-06-25
Add argoCD and argoCD apps to vcluster

iamvigneshwars:


charts/argocd-apps/Chart.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v2
+name: argocd
```
garryod, line 2: Might be worth thinking about the naming of this and the directory, could maybe just go with `apps`?

charts/argocd-apps/templates/events-application.yaml
garryod, line 1: Can we add an `{{- if .Values.events.enabled -}}` to this and the others? Would be good to be able to turn apps on add off easily

charts/argocd-apps/templates/events-application.yaml
```diff
@@ -0,0 +1,23 @@
...
+  namespace: argocd
+spec:
+  destination:
+    namespace: events
+    server: {{ .Values.spec.destination.server }}
+  project: default
+  source:
+    helm:
+      valueFiles:
+      - values.yaml
```
garryod, line 14: Could we use a `{{- if .Values.dev.enabled -}}` to use the `dev-values` for each app?

garryod (changes requested):

charts/argocd-apps/templates/events-application.yaml
```diff
@@ -0,0 +1,29 @@
...
+    server: {{ .Values.spec.destination.server }}
+  project: default
+  source:
+    helm:
+      valueFiles:
+      {{- if.Values.dev.enabled }}
+      - dev-values.yaml
+      {{- else }}
+      - values.yaml
+      {{- end }}
```
garryod, line 19: We should always have the `values.yaml` with `dev-values.yaml` overriding, not sure which order is correct though
```suggestion
      {{- if.Values.dev.enabled }}
      - dev-values.yaml
      {{- end }}
      - values.yaml
```

garryod (changes requested):

garryod (approved):

PR #46, opened 2024-06-28
Getting the vcluster/argocd deployment working in dev mode

daurer:
- [x] deployed app of apps with argocd as a bundle
- [x] downsized resources in dev mode


charts/workflows-cluster/dev-values.yaml
```diff
@@ -47,13 +47,101 @@ vcluster:
...
+                requests:
+                  cpu: 10m
+                  memory: 256Mi
+        - release:
+            name: argocd-app-of-apps
             namespace: argocd
+          # This is an empty helm chart compressed into a base64 string
+          # The actual content is defined in the values below
+          # The reason for this hack is that helm charts and manifests
+          # are not executed in order (should be fixed upstream in vcluster)
```
garryod, line 121:
```suggestion
          # This is the `charts/argocd-app-of-apps` chart as a packaged base64 string.
          # You can re-create it with `helm package charts/argocd-app-of-apps && base64 aoa-*.tgz`.
          # This hack allows us to deploy manifests in a deterministic order, which is not possible using `manifests`.
```

charts/argocd-app-of-apps/Chart.yaml
garryod, line 1: Could we call this something that's descriptive of what the chart actually does? Perhaps `verbatim-values`?

garryod (changes requested):

charts/argocd-app-of-apps/Chart.yaml
daurer: `verflixt` seems like a good name :1st_place_medal:

PR #59, opened 2024-07-08
Added intro and how-to on submitting a workflow

daurer:
- [x] Write a quick intro
- [x] Add how-to on how to submit a workflow
- [x] Explain how to submit with dashboard

### DIFF
--- a/docs/how-tos/submit-workflow.md
+++ b/docs/how-tos/submit-workflow.md
@@ -3,7 +3,8 @@
 ## Preface
 
 This guide will describe how to configure and submit a workflow that is then running in the namespace of a Diamond visit.
-It is recommended to build a container for your application and push it to a container registry.
+The easiest way to run your application as a workflow is by using containers.
+We therefore recommmend that you build a container for your application and push it to a container registry.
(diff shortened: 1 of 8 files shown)


docs/how-tos/submit-workflow.md
```diff
@@ -0,0 +1,87 @@
...
+We can then provide a message at runtime as a parameter when we submit the workflow.
+
+## Using the Argo dashboard to submit a workflow
+
+
+## Using the Argo CLI to submit a workflow
+
+!!! info "Authentication"
+
+    Since we currently rely on CAS for authentication, we cannot use the Argo CLI to submit a workflow.
```
garryod, line 86:
```suggestion
!!! warn "Authentication"

    Since we currently rely on CAS for authentication, we cannot use the Argo CLI to submit a workflow.
```

docs/how-tos/submit-workflow.md
garryod, line 1: Feel like we should probably link out to the Argo workflows docs somewhere in here, perhaps in multiple places even

garryod (approved):

PR #67, opened 2024-07-10
Disable argo-events app

garryod:
This is not currently production ready, therefore it should be disabled in the app of apps


daurer (approved):
make sense

PR #73, opened 2024-07-10
Setup sync waves for argocd apps

garryod:


daurer (approved):
Looks fine, we should probably double check if the syncing actually happens in this order

PR #75, opened 2024-07-10
Template target revision in app of apps

garryod:


daurer (commented):
There should probably also be a default that is  set in the values/dev-values of argocd-apps?

garryod:
> There should probably also be a default that is set in the values/dev-values of argocd-apps?

Opted to put a `targetRevision` in each of the values entries, tempted to revisit the `dev` area in the future

daurer (approved):

PR #79, opened 2024-07-11
Create group ArgoCD projects and apps

garryod:
Adds a helm chart which:
- Creates a namespace for each group (with the name `<NAME>-group`)
- Creates an ArgoCD [AppProject](https://argo-cd.readthedocs.io/en/stable/user-guide/projects/) for each group, which can only deploy (Cluster)WorkflowTemplates
- Creates a set of Apps for each group

### DIFF
--- a/charts/argocd-apps/values.yaml
+++ b/charts/argocd-apps/values.yaml
@@ -7,6 +7,12 @@ events:
   extraValuesFiles: []
   valuesObject: {}
 
+groups:
+  enabled: true
+  targetRevision: HEAD
+  extraValuesFiles: []
+  valuesObject: {}
+
 kyverno:
   enabled: true
   targetRevision: 3.2.2
--- a/charts/groups/templates/namespaces.yaml
+++ b/charts/groups/templates/namespaces.yaml
@@ -1,0 +1,8 @@
+{{- range .Values.groups }}
+{{- if .enabled }}
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: "{{ .name }}-group"
+{{- end }}
+{{- end }}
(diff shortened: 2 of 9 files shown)


charts/groups/values.yaml
```diff
@@ -0,0 +1,14 @@
...
+  server: ""
+
+groups:
+  - name: imaging
+    enabled: true
+    permittedRepos:
+      - https://github.com/DiamondLightSource/imaging-workflows
+    apps:
+      - name: imaging
+        repoURL: https://github.com/DiamondLightSource/imaging-workflows
```
daurer, line 12: can we somehow avoid listing all the repos twice?

charts/argocd-apps/values.yaml
```diff
@@ -64,6 +70,6 @@ sealedsecrets:
 
 workflows:
   enabled: true
-  targetRevision: HEAD
+  targetRevision: app-projects
```
daurer, line 73: don't forget to clean up before final merge

charts/groups/templates/namespaces.yaml
daurer, line 1: what do we need this namespace for?

charts/groups/values.yaml
garryod: Yeah, should be an easy-ish update to the chart. Though maybe we don't even want permitted repos and should just allow them to use whatever?

charts/groups/templates/namespaces.yaml
garryod: Don't absolutely need it, but it makes it easier to write kyverno policy which equates to ArgoCD Signals in the `imaging-group` namespace can create workflows in namespaces x, y & z. Or to give folks from the group permission to modify resources in the namespace

daurer (approved):

PR #84, opened 2024-07-15
Fix workflows default-user permissions

garryod:
The `apiGroup` of the `roleRef` in the ClusterRoleBinding for the `default-user` ServiceAccount was missing the `/v1` at the end of `rbac.authorization.k8s.io/v1`, hence the role was not bound. Also tidied some stuff up whilst trying to find the mistake :sweat_smile:


charts/workflows/templates/visit-member-clusterrole.yaml
```diff
@@ -3,18 +3,24 @@ kind: ClusterRole
...
+    verbs:
+      - get
+      - watch
+      - list
+  - apiGroups:
+      - argoproj.io
+    resources:
+      - workflows
+    verbs:
+      - create
```
daurer, line 26: might wanna let visit member also delete workflows

daurer (approved):

PR #93, opened 2024-07-25
Reduce kyverno generation cycling

garryod:
Kyverno is applying policies in an infinite loop, there appears to be several underlying issues...
- Kyverno `1.12.3` was duplicating UpdateRequests for background policies ([PR](https://github.com/kyverno/kyverno/pull/10431)), so reverted kyverno to `1.10.7`
- [Kyverno docs](https://kyverno.io/docs/writing-policies/tips/#generate) suggests that generate rules which are triggered by the same resource reside in the same policy, so all were merged
- Generation was slow, so increased `genWorkers` to 20
- Kyverno reports & cleanup controller were unused, so disabled them


daurer (approved):

daurer:
maybe wait with the merge until we know that all the policies have been applied as expected

daurer (dismissed):

daurer (approved):

PR #108, opened 2024-08-02
Added tutorial for ptypy workflow

daurer:
Added a tutorial describing how to create a cluster workflow template for ptypy, specificially how to

- containerise the application
- write a template and make it visible to workflow controller
- add parameters as input arguments
- mount the visit directory to /dls
- specify cluster resources (CPU and GPU)
- deal with non-root containers

### DIFF
--- a/docs/tutorials/ptypy-workflows.md
+++ b/docs/tutorials/ptypy-workflows.md
@@ -9,7 +9,7 @@
 ## Containerising the Application
 
 The first step is to build a [container](https://dev-portal.diamond.ac.uk/guide/kubernetes/tutorials/containers/)
-for PtyPy, for example the following multi-stage Dockerfile
+for PtyPy, for example the following multi-stage Dockerfile in the box below
 
 ??? example "Dockerfile"
     ```dockerfile
@@ -88,13 +88,28 @@ for PtyPy, for example the following multi-stage Dockerfile
     # Run PtyPy run script as entrypoint
     ENTRYPOINT ["ptypy.run"]
     ```
-can be used to build a container image with openmpi as the MPI backend and CuPy as the backend for PtyPy
+
+### Building a container
+
+To build a container image `ptypy-container` with openmpi as the MPI backend and CuPy as the backend for PtyPy, 
+we can use the following podman command
 
 ```bash
-podman build . -t gcr.io/diamond-privreg/ptypy/test_openmpi_cupy:0.1 --target=runtime --build-arg MPI=openmpi --build-arg PLATFORM=cupy
+podman build . -t ptypy-container --target=runtime --build-arg MPI=openmpi --build-arg PLATFORM=cupy
 ```
 
-and then be pushed to Diamond's the private container registry
+### Publishing a container
+
+If we wish to publish our container, we could do so by pushing it to
+Diamond's container registry on the Google cloud,
+see [here](https://dev-portal.diamond.ac.uk/guide/kubernetes/tutorials/containers/) details on how to set this up.
+Now we can change the tag of our image
+
+```bash
+podman tag ptypy-container gcr.io/diamond-privreg/ptypy/test_openmpi_cupy:0.1
+```
+
+and push
 
 ```bash
 podman push gcr.io/diamond-privreg/ptypy/test_openmpi_cupy:0.1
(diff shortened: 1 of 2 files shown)


docs/tutorials/ptypy-workflows.md
```diff
@@ -0,0 +1,440 @@
...
+    FROM build as cupy-post
+    RUN pip install ./cufft
+
+    # Platform specific runtime container
+    FROM ${PLATFORM}-post as runtime
+
+    # Run PtyPy run script as entrypoint
+    ENTRYPOINT ["ptypy.run"]
+    ```
+can be used to build a container image with openmpi as the MPI backend and CuPy as the backend for PtyPy
```
garryod, line 91: Probably best off not splitting this paragraph with an admonition

docs/tutorials/ptypy-workflows.md
```diff
@@ -0,0 +1,440 @@
...
+
+```bash
+podman build . -t gcr.io/diamond-privreg/ptypy/test_openmpi_cupy:0.1 --target=runtime --build-arg MPI=openmpi --build-arg PLATFORM=cupy
+```
+
+and then be pushed to Diamond's the private container registry
+
+```bash
+podman push gcr.io/diamond-privreg/ptypy/test_openmpi_cupy:0.1
+```
```
garryod, line 101: It would be perhaps best to link out to pushing to `diamond-{privreg,pubreg}` docs, as there is a bit of setup required and they're due to be replaced with GitLab container registry in the near future

docs/tutorials/ptypy-workflows.md
```diff
@@ -0,0 +1,440 @@
...
+
+## Cluster Workflow Templates
+
+The next step is to create a ```ClusterWorkFlowTemplate``` and push it to the corresponding GitHub
+repository for science workflows, in this case [imaging-workflows](https://github.com/DiamondLightSource/imaging-workflows).
+The Argo workflow engine can be configured to automatically look for existing templates in those
+repositories and make them available via the [dashboard](https://workflows.diamond.ac.uk) or the graph (once available).
```
garryod, line 108: This is a bit of a mis-truth, so probably worth re-wording to make clear the separation between ArgoCD and Argo Workflows

docs/tutorials/ptypy-workflows.md
```diff
@@ -0,0 +1,440 @@
...
+  name: ptypy-cpu-job
+spec:
+  entrypoint: ptypy-run
+  templates:
+  - name: ptypy-run
+    container:
+      image: gcr.io/diamond-privreg/ptypy/test_openmpi_full:0.1
+```
+
+defining a ClusterWorkflowTemplate resource with a given name (```ptypy-cpu-job```) and a list of tasks. Here, we
```
garryod, line 133:
```suggestion
defining a `ClusterWorkflowTemplate` resource with a given name (`ptypy-cpu-job`) and a list of tasks. Here, we
```

docs/tutorials/ptypy-workflows.md
```diff
@@ -0,0 +1,440 @@
...
+## Writing a Cluster Workflow Template
+
+Cluster Workflow Templates are largely the same as [Workflow Templates](../how-tos/submit-workflow.md)
+with a basic structure looking like this
+
+```yaml
+apiVersion: argoproj.io/v1alpha1
+kind: ClusterWorkflowTemplate
+metadata:
+  name: ptypy-cpu-job
```
garryod, line 124: Perhaps better to use `generateName` here?

docs/tutorials/ptypy-workflows.md
```diff
@@ -0,0 +1,440 @@
...
+      effect: NoSchedule
+```
+
+### Non-root container execution
+
+For security reasons, workflow pods are only allowed to run with non-root users,
+this means that the application running inside the container won't have write access
+to anything except mounted directories.
+Since many applications will need to write to some locations such as ```/tmp```,
+it makes sense to add a volume claim for a temporary directory to our template
```
garryod, line 275: Probably better to phrase this as "if required" - a lot of containers won't need this

garryod (changes requested):
Looking good, hopefully in the future we can have each section link out to a full how-to guide or explanation

docs/tutorials/ptypy-workflows.md
daurer: Added additional subsections to avoid splitting having Dockerfile in the middle of a paragraph

docs/tutorials/ptypy-workflows.md
daurer: Included new subsection on publishing to a registry and made explicit link to instructions on how to set it up.

docs/tutorials/ptypy-workflows.md
daurer: Thanks for catching this. Changed text to make clarify the difference between ArgoCD and Argo workflows.

docs/tutorials/ptypy-workflows.md
daurer: not sure if `generateName` was working in a ClusterWorkflowTemplate. Can check once all the CRDs are available

docs/tutorials/ptypy-workflows.md
daurer: Done a bit of rewriting to make clear that this is optional and might not be needed.

garryod (approved):

PR #135, opened 2024-08-22
Setup `graph-proxy` service

garryod:


iamvigneshwars (changes requested):
Looks Great! has few EOF's and misspelled words!

iamvigneshwars (approved):
LGTM

PR #140, opened 2024-08-27
Add TasksFlow component

iamvigneshwars:
This PR adds a react component that takes a list of workflow tasks and create a Directed Acyclic Graph. Tests checks if the component uses the props and functions properly and renders without crashing, also includes tests for the custom functions used by the component.

### DIFF
--- a/.github/workflows/frontend.yaml
+++ b/.github/workflows/frontend.yaml
@@ -47,4 +47,4 @@ jobs:
         run: yarn --frozen-lockfile
 
       - name: Test
-        run: yarn test --passWithNoTests # Currently there are no tests
+        run: yarn test --verbose
--- a/frontend/jest.config.cjs
+++ b/frontend/jest.config.cjs
@@ -7,7 +7,9 @@ module.exports = {
     "**/?(*.)+(spec|test).[jt]s?(x)",
   ],
   moduleNameMapper: {
-    "^@workflows-lib/(.*)$": "workflows-lib/src/$1",
-    "^@relay-workflows-lib/(.*)$": "relay-workflows-lib/src/$1",
+    "^@workflows-lib/(.*)$": "workflows-lib/lib/$1",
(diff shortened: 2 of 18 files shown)


frontend/workflows-lib/lib/components/workflow/CustomNode.tsx
garryod, line 1: Might be worth naming this something task-y

frontend/workflows-lib/lib/uilts/TasksFlowUtils.ts
```diff
@@ -0,0 +1,89 @@
+import dagre from "@dagrejs/dagre";
+import { Node, Edge } from "@xyflow/react";
+import { Task, TaskNode } from "../types";
+
+export function applyDagreLayout(
+  nodes: Node[],
+  edges: Edge[]
+): { nodes: Node[]; edges: Edge[] } {
+  const g = new dagre.graphlib.Graph();
```
garryod, line 9: Can we rename `g` to `graph`

frontend/workflows-lib/lib/uilts/TasksFlowUtils.ts
garryod, line 1: Could we move these methods to a file close to the workflow component?

frontend/workflows-lib/stories/mockData/MockTasks.ts
garryod, line 1: Can we just dump this data in the story?

garryod (changes requested):
Got a few small things, but looking good
@abigailalexander would you mind taking a quick look?

garryod (approved):
LGTM

garryod (changes requested):
Any chance you could refactor the commits before merging? Mainly just to get rid of the last commit

abigailalexander (approved):
LGTM as well

garryod (approved):
LGTM

PR #149, opened 2024-08-29
Add Tasks Table component

iamvigneshwars:


frontend/workflows-lib/lib/components/workflow/TasksTable.tsx
```diff
@@ -0,0 +1,40 @@
...
+import { getStatusIcon } from "../common/StatusIcons";
+
+interface TaskTableProps {
+  tasks: Task[];
+}
+
+const TasksTable: React.FC<TaskTableProps> = ({ tasks }) => {
+  const theme = useTheme();
+  return (
+    <Box display="flex" flexWrap="wrap" gap={1.5} padding={2} overflow="auto">
```
garryod, line 13: Would we be better off using a `grid` here?

frontend/workflows-lib/lib/components/workflow/TasksTable.tsx
```diff
@@ -0,0 +1,40 @@
...
+            minWidth: 180,
+            maxWidth: 180,
+          }}
+        >
+          <Box
+            display="flex"
+            justifyContent="space-between"
+            alignItems="center"
+            minWidth={120}
+          >
```
garryod, line 28: Do we need a `box` inside the `paper` or could we style the paper to make it work without?

garryod (approved):
LGTM

PR #158, opened 2024-08-30
Add dynamic tasks component

iamvigneshwars:

PR #167, opened 2024-09-09
add Workflow component

iamvigneshwars:
depends #158

### DIFF
--- a/frontend/workflows-lib/lib/types.ts
+++ b/frontend/workflows-lib/lib/types.ts
@@ -8,3 +8,16 @@ export interface Task {
 export interface TaskNode extends Task {
   children?: TaskNode[];
 }
+
+export type WorkflowStatus =
+  | "Pending"
+  | "Running"
+  | "Succeeded"
+  | "Failed"
+  | "Errored";
+
+export interface Workflow {
+  name: string;
+  status: WorkflowStatus;
+  tasks: Task[];
+}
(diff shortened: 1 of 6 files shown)


frontend/workflows-lib/stories/WorkflowAccordion.stories.tsx
```diff
@@ -0,0 +1,95 @@
+import type { Meta, StoryObj } from "@storybook/react";
+import WorkflowAccordion from "../lib/components/workflow/WorkflowAccordian";
+import { WorkflowStatus } from "../lib/types";
+
+const fakeTasks = [
```
garryod, line 5: Could we move these to a common TS / JSON file?

frontend/workflows-lib/lib/types.ts
```diff
@@ -8,3 +8,17 @@ export interface Task {
...
   children?: TaskNode[];
 }
+
+export enum WorkflowStatus {
+  Pending = "Pending",
+  Running = "Running",
+  Succeeded = "Succeeded",
+  Failed = "Failed",
+  Errored = "Errored",
+}
```
garryod, line 18: Would it be better to use a union type (e.g. `type WorkflowStatus = "Pending" | "Running"`) instead?

frontend/workflows-lib/stories/WorkflowAccordion.stories.tsx
iamvigneshwars: That could be a separate PR? we could move it for all stories?

garryod (approved):
LGTM

PR #178, opened 2024-09-19
fix argo workflow API schema

iamvigneshwars:
The argo workflows API schema does not match with its API response. This is a temporary fix to the schema.

### DIFF
--- a/graph-proxy/argo-workflows-openapi/build.rs
+++ b/graph-proxy/argo-workflows-openapi/build.rs
@@ -1,4 +1,5 @@
 use quote::quote;
+use schemars::schema::{InstanceType, RootSchema, Schema, SchemaObject};
 use std::{
     env,
     fs::{create_dir_all, File},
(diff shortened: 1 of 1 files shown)


graph-proxy/argo-workflows-openapi/build.rs
```diff
@@ -12,7 +13,23 @@ fn main() {
         .unwrap()
         .text()
         .unwrap();
-    let schema = serde_json::from_str::<schemars::schema::RootSchema>(&raw_schema).unwrap();
+    let mut schema: RootSchema = serde_json::from_str(&raw_schema).unwrap();
+
+    if let Some(Schema::Object(ref mut pod_gc)) = schema
```
garryod, line 18: Could we add a comment explaining why this is done with a link to the upstream issue?

graph-proxy/argo-workflows-openapi/build.rs
```diff
@@ -12,7 +13,23 @@ fn main() {
...
+        .definitions
+        .get_mut("io.argoproj.workflow.v1alpha1.PodGC")
+    {
+        if let Some(pod_gc_object) = &mut pod_gc.object {
```
garryod, line 22: Should this fail silently if the structure changes? Or should we just `unwrap`?

garryod (changes requested):

garryod (approved):

PR #183, opened 2024-09-23
add tasks to graphql proxy

iamvigneshwars:
Depends on #178

### DIFF
--- a/graph-proxy/src/graphql/workflows.rs
+++ b/graph-proxy/src/graphql/workflows.rs
@@ -1,11 +1,12 @@
 use crate::ArgoServerUrl;
 use argo_workflows_openapi::{
-    APIResult, IoArgoprojWorkflowV1alpha1Workflow, IoArgoprojWorkflowV1alpha1WorkflowStatus,
+    APIResult, IoArgoprojWorkflowV1alpha1NodeStatus, IoArgoprojWorkflowV1alpha1Workflow,
+    IoArgoprojWorkflowV1alpha1WorkflowStatus,
 };
-use async_graphql::{Context, Object, SimpleObject, Union};
+use async_graphql::{Context, Enum, Object, SimpleObject, Union};
 use axum_extra::headers::{authorization::Bearer, Authorization};
 use chrono::{DateTime, Utc};
-use std::ops::Deref;
+use std::{collections::HashMap, ops::Deref};
 use tracing::{debug, instrument};
 
 /// An error encountered when parsing the Argo Server API Workflow response
@@ -18,6 +19,10 @@ enum WorkflowParsingError {
     MissingStartTime,
     #[error("status.end_time was expected but was not present")]
     MissingEndTime,
+    #[error("value.phase was not a recognised value")]
+    UnrecognisedTaskPhase,
+    #[error("value.display_name was not a recognised value")]
+    UnrecognisedTaskDisplayName,
 }
 
 /// A Workflow consisting of one or more [`Task`]s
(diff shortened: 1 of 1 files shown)


graph-proxy/src/graphql/workflows.rs
```diff
@@ -185,6 +189,114 @@ impl TryFrom<IoArgoprojWorkflowV1alpha1WorkflowStatus> for WorkflowErroredStatus
...
+    /// Display name of the task
+    name: String,
+    /// Current status of a task
+    status: TaskStatus,
+    /// Parent of a task
+    depends: Vec<String>,
+}
+
+/// A wrapper for list of tasks
+struct TaskMap(HashMap<String, IoArgoprojWorkflowV1alpha1NodeStatus>);
```
garryod, line 235: Do we want a new type here or just a type alias?

graph-proxy/src/graphql/workflows.rs
```diff
@@ -185,6 +189,114 @@ impl TryFrom<IoArgoprojWorkflowV1alpha1WorkflowStatus> for WorkflowErroredStatus
...
+        let relationship_map = value.generate_relationship_map();
+
+        let tasks = value
+            .0
+            .into_iter()
+            .map(|(node_name, node_status)| {
+                let mut task = Task::try_from(node_status)?;
+                task.depends = relationship_map
+                    .get(&node_name)
+                    .cloned()
```
garryod, line 290: Since each is only got once, we can use `.remove(&node_name).unwrap_or_default()` in place of `.get(&node_name).cloned().unwrap_or_default()`

graph-proxy/src/graphql/workflows.rs
```diff
@@ -185,6 +189,114 @@ impl TryFrom<IoArgoprojWorkflowV1alpha1WorkflowStatus> for WorkflowErroredStatus
...
+    type Error = WorkflowParsingError;
+
+    fn try_from(value: TaskMap) -> Result<Self, Self::Error> {
+        let relationship_map = value.generate_relationship_map();
+
+        let tasks = value
+            .0
+            .into_iter()
+            .map(|(node_name, node_status)| {
+                let mut task = Task::try_from(node_status)?;
```
garryod, line 287: Instead of creating from the `node_status` then mutating with the dependencies, could we create a single function which takes both and produces a task?

graph-proxy/src/graphql/workflows.rs
```diff
@@ -185,6 +189,114 @@ impl TryFrom<IoArgoprojWorkflowV1alpha1WorkflowStatus> for WorkflowErroredStatus
...
+#[derive(Debug, SimpleObject)]
+struct Task {
+    /// Unique name of the task
+    id: String,
+    /// Display name of the task
+    name: String,
+    /// Current status of a task
+    status: TaskStatus,
+    /// Parent of a task
+    depends: Vec<String>,
```
garryod, line 231: Is it worth also providing dependencies?

graph-proxy/src/graphql/workflows.rs
```diff
@@ -185,6 +189,114 @@ impl TryFrom<IoArgoprojWorkflowV1alpha1WorkflowStatus> for WorkflowErroredStatus
...
+            .into_iter()
+            .map(|(node_name, node_status)| {
+                let mut task = Task::try_from(node_status)?;
+                task.depends = relationship_map
+                    .get(&node_name)
+                    .cloned()
+                    .unwrap_or_default();
+                Ok(task)
+            })
+            .collect::<Result<Vec<Task>, Self::Error>>()?;
```
garryod, line 294: Can we infer the types here? e.g. `collect::<Result<Vec<_>, _>>`?

garryod (changes requested):

graph-proxy/src/graphql/workflows.rs
iamvigneshwars: I have custom methods, so we need struct? 🤔

garryod: Yeah, seems reasonable. Other option would be to just inline the dependency resolution into the `TryFrom`

iamvigneshwars: But we also have another custom method `generate_relationship_map`?  is it gonna make much difference? is't this easier to work with?

garryod (approved):

PR #185, opened 2024-09-23
add Relay workflow component

iamvigneshwars:


garryod (changes requested):
Since we have to make additional requests to the Argo Workflows Server to retrieve tasks, would it perhaps make sense to ship our component with `Tasks` hidden by default? Then we can make an additional request to retrieve the task list if a user opens it

PR #190, opened 2024-09-25
add list of workflows to graph proxy 

iamvigneshwars:


graph-proxy/src/graphql/workflows.rs
```diff
@@ -347,4 +356,75 @@ impl WorkflowsQuery {
...
+
+    #[instrument(skip(self, ctx))]
+    async fn workflows(
+        &self,
+        ctx: &Context<'_>,
+        proposal_code: String,
+        proposal_number: u32,
+        visit: u32,
+        _continue: Option<String>,
+        limit: Option<u32>,
```
garryod, line 368: We want to follow the [GraphQL cursor connections spec](https://relay.dev/graphql/connections.htm). `async-graphql` has [some tooling](https://async-graphql.github.io/async-graphql/en/cursor_connections.html) for this

graph-proxy/src/graphql/workflows.rs
```diff
@@ -347,4 +356,75 @@ impl WorkflowsQuery {
...
+                .append_pair("listOptions.continue", &cursor);
+        }
+        debug!("Retrieving workflows name from {url}");
+        let workflows_response = request(url)
+            .send()
+            .await?
+            .json::<APIResult<argo_workflows_openapi::IoArgoprojWorkflowV1alpha1WorkflowList>>()
+            .await?
+            .into_result()?;
+        let workflows =
```
garryod, line 399: This should be done lazily so if someone just requests the `name` then we don't have to perform ten extra requests

graph-proxy/src/graphql/workflows.rs
```diff
@@ -347,4 +356,75 @@ impl WorkflowsQuery {
...
+            future::try_join_all(workflows_response.items.into_iter().filter_map(|workflow| {
```
garryod, line 400: Can we use a [`tokio::task::JoinSet`](https://docs.rs/tokio/latest/tokio/task/struct.JoinSet.html) here instead?

garryod (changes requested):

graph-proxy/src/graphql/workflows.rs
```diff
@@ -347,4 +347,118 @@ impl WorkflowsQuery {
...
+                client.get(url).bearer_auth(auth_token.token())
+            } else {
+                client.get(url)
+            }
+        };
+        let mut url = server_url.clone();
+        url.path_segments_mut()
+            .unwrap()
+            .extend(["api", "v1", "workflows", &namespace]);
+        let limit = limit.unwrap_or(10).min(10);
```
garryod, line 376: Could we use an [input value validator instead](https://async-graphql.github.io/async-graphql/en/input_value_validators.html)? Think they automagically provide good error messages

graph-proxy/src/graphql/workflows.rs
```diff
@@ -347,4 +347,118 @@ impl WorkflowsQuery {
...
+        url.query_pairs_mut()
+            .append_pair("listOptions.limit", &limit.to_string());
+        let current_cursor = cursor
+            .clone()
```
garryod, line 380: Do we need to clone here? We take a reference to it inside the `map`

graph-proxy/src/graphql/workflows.rs
```diff
@@ -347,4 +347,118 @@ impl WorkflowsQuery {
...
+            .map(|cursor| {
+                OpaqueCursor::<String>::decode_cursor(&cursor)
+                    .map(|cursor_value| {
+                        url.query_pairs_mut()
+                            .append_pair("listOptions.continue", &cursor_value.0);
+                        cursor_value.0
+                    })
```
garryod, line 387: Performing a mutation inside a map is generally discouraged as they're lazily evaluated. Could we instead just map to the decoded cursor and use an `if` to perform the mutation

graph-proxy/src/graphql/workflows.rs
```diff
@@ -347,4 +347,118 @@ impl WorkflowsQuery {
...
+                    .map_err(|_| anyhow::Error::msg("Cursor not valid"))
```
garryod, line 388: Is there a reason the error returned by `decode_cursor` can't be shortcut itself? Or with a `anyhow::Context`

garryod (commented):
Looking pretty good, would be nice to see the lazy version of this before we decide on which to go with

graph-proxy/src/graphql/workflows.rs
```diff
@@ -309,6 +312,295 @@ impl TryFrom<TaskMap> for Vec<Task> {
...
+            status: LazyWorkflowStatus::try_from((value.status.unwrap(), metadata))?,
+        })
+    }
+}
+lazy_static! {
+    static ref CLIENT: reqwest::Client = reqwest::Client::new();
+}
+#[allow(clippy::missing_docs_in_private_items)]
+macro_rules! fetch_tasks {
+    ($namespace:expr, $workflow_name:expr, $ctx:expr) => {{
```
garryod, line 352: Think we could have this macro also create the function definition - just to compact things a little more

graph-proxy/src/graphql/workflows.rs
garryod, line 1: Instead of creating a `LazyWorkflow` and `LazyWorkflow{Pending | Running | Succeeded | Failed | Errored}Status`, could we simply modify the original `Workflow` and it's statuses to act in this way?

Would mean storing the `tasks` as an `Option<Vec<Task>>` and having `from IoArgoprojWorkflowV1alpha1WorkflowStatus for Workflow` create `Some` whilst  from `from IoArgoprojWorkflowV1alpha1WorkflowList for Workflow` would create `None` with the `tasks` resolver then performing a query to fill in`tasks` if it was `None`

graph-proxy/src/graphql/workflows.rs
```diff
@@ -56,22 +81,70 @@ enum WorkflowStatus {
...
     }
 }
 
+lazy_static! {
+    static ref CLIENT: reqwest::Client = reqwest::Client::new();
+}
+
+#[allow(clippy::missing_docs_in_private_items)]
+macro_rules! fetch_tasks {
+    ($status:expr, $ctx:expr) => {{
```
garryod, line 110: Since we're not generating the function declaration I think we can just use a function here. Alternatively we could create a macro to generate the entire `impl $iden`

garryod (changes requested):
Looking good, just a few bits and pieces that could be tidied up

garryod (commented):
Looking good, just a few bits that can be tidied up

graph-proxy/src/graphql/workflows.rs
```diff
@@ -345,6 +438,75 @@ impl WorkflowsQuery {
...
+    }
+
+    #[instrument(skip(self, ctx))]
+    async fn workflows(
+        &self,
+        ctx: &Context<'_>,
+        proposal_code: String,
+        proposal_number: u32,
+        visit: u32,
+        cursor: Option<String>,
```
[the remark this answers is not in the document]
garryod: Gotcha

graph-proxy/src/graphql/workflows.rs
```diff
@@ -56,22 +78,30 @@ enum WorkflowStatus {
     Errored(WorkflowErroredStatus),
 }
 
-impl TryFrom<IoArgoprojWorkflowV1alpha1WorkflowStatus> for WorkflowStatus {
-    type Error = WorkflowParsingError;
-
-    fn try_from(value: IoArgoprojWorkflowV1alpha1WorkflowStatus) -> Result<Self, Self::Error> {
+#[allow(clippy::missing_docs_in_private_items)]
```
garryod, line 81: Docs would be good here

graph-proxy/src/graphql/workflows.rs
```diff
@@ -270,6 +313,78 @@ impl Task {
     }
 }
 
+#[allow(clippy::missing_docs_in_private_items)]
+#[derive(Debug)]
+enum Tasks {
+    Fetched(Vec<Task>),
+    UnFetched(Arc<Metadata>),
+}
```
garryod, line 321: Missing whitespace

graph-proxy/src/graphql/workflows.rs
```diff
@@ -270,6 +313,78 @@ impl Task {
...
+        }
+    }
+}
+
+impl Tasks {
+    /// Converts [`TaskMap`] into [`Tasks`]`
+    fn into_tasks(
+        taskmap: TaskMap,
+        metadata: Arc<Metadata>,
+    ) -> Result<Tasks, WorkflowParsingError> {
```
garryod, line 371: Surely this should be implemented on `TaskMap` using `self` as the first parameter? Or be called `from_taskmap`

iamvigneshwars: "Would probably make more sense to implement this on `Tasks`" this was your previous comment when I did that 😢 😩

garryod: Ah yeah, didn't see it took `metadata` too. Definitely have it on `Tasks` and call it `from_taskmap` or something similar

garryod: But `Tasks::into_tasks` should have the signature `fn into_tasks(self: Tasks) -> Tasks`, which would be a bit silly

garryod (approved):
LGTM

PR #199, opened 2024-10-02
Update session namespace explanation

garryod:
This is out of date following the transition to KeyCloak and Argo Workflows Client Authentication


daurer (approved):
LGTM

PR #202, opened 2024-10-03
update workflow component to take any child component as prop

iamvigneshwars:


frontend/workflows-lib/lib/components/workflow/WorkflowAccordian.tsx
```diff
@@ -15,7 +15,10 @@ const AccordionDetails = styled(MuiAccordionDetails)(({ theme }) => ({
   height: "80vh",
 }));
 
-const WorkflowAccordion: React.FC<WorkflowProps> = ({ workflow }) => {
+const WorkflowAccordion: React.FC<WorkflowProps> = ({
+  workflow,
+  ChildComponent,
+}) => {
```
garryod, line 21: Is there a reason we're passing it like this an not [as `children`](https://react.dev/learn/passing-props-to-a-component#passing-jsx-as-children)?

garryod (commented):
Looking good, just one question on `children`

frontend/workflows-lib/lib/components/workflow/WorkflowAccordian.tsx
iamvigneshwars: We're doing the same thing, we just have another prop for the workflow name and status.

garryod: Yeah, but if we were to call it `children` then we could use the normal JSX syntax of:
```javascript
<WorkflowAccordion>
    <Tasks/>
</WorkflowAccordion>
```

iamvigneshwars: Ahh got it, fixed it!

PR #204, opened 2024-10-04
Fixed further issues when deploying in dev mode

daurer:


charts/argocd-apps/values.yaml
```diff
@@ -2,7 +2,7 @@ destination:
   server: https://kubernetes.default.svc
 
 events:
-  enabled: false
+  enabled: true
```
garryod, line 5: Is this production ready now?

daurer: Good catch, I'd say probably not...

garryod (dismissed):
LGTM

garryod (approved):
LGTM

PR #208, opened 2024-10-08
add Relay workflow component

iamvigneshwars:


.github/workflows/frontend.yaml
```diff
@@ -24,6 +24,22 @@ jobs:
       - name: Install dependencies
         run: yarn --frozen-lockfile
 
+      - name: Download graph-federation tar.gz
+        run: wget https://github.com/DiamondLightSource/graph-federation/archive/refs/tags/1.0.0.tar.gz -O graph-federation.tar.gz
```
garryod, line 28: We don't want to pin to a tag here. Otherwise we won't catch issues caused by updating the schema. Might be just worth building the crate and getting the schema that way :thinking:

.github/workflows/frontend.yaml
```diff
@@ -24,6 +24,22 @@ jobs:
...
+
+      - name: Extract graph-federation archive
+        run: tar -xzf graph-federation.tar.gz
+
+      - name: Copy workflows.graphql to relay-workflows-lib
+        run: cp graph-federation-1.0.0/schema/workflows.graphql relay-workflows-lib/
+
+      - name: Remove directives from workflows.graphql
+        run: sed -i -e :a -e '$d;N;2,8ba' -e 'P;D' relay-workflows-lib/workflows.graphql
```
garryod, line 37: Why do we need to remove the directives?

frontend/relay-workflows-lib/lib/components/WorkflowRelay.tsx
```diff
@@ -0,0 +1,88 @@
...
+  const statusTextMap: { [key: string]: WorkflowStatus } = {
+    WorkflowPendingStatus: "Pending",
+    WorkflowRunningStatus: "Running",
+    WorkflowSucceededStatus: "Succeeded",
+    WorkflowFailedStatus: "Failed",
+    WorkflowErroredStatus: "Errored",
+  };
+  const statusText = statusTextMap[data.status.__typename];
+  return (
+    <div>
```
garryod, line 75: Why the `<div>`?

frontend/relay-workflows-lib/lib/components/WorkflowRelay.tsx
```diff
@@ -0,0 +1,88 @@
...
+import { graphql } from "react-relay";
+import { WorkflowAccordion } from "workflows-lib";
+import type { WorkflowStatus } from "workflows-lib";
+import { WorkflowRelayFragment$key } from "./__generated__/WorkflowRelayFragment.graphql";
+
+const WorkflowRelay = (props: {
+  workflow: WorkflowRelayFragment$key;
+  children: React.ReactNode;
+}) => {
+  const workflowFragment = graphql`
```
garryod, line 11: Can we remove a lot of the fields from this query? Doesn't seem like any of the `status { ... on }` clauses are required here

garryod (changes requested):

.github/workflows/frontend.yaml
iamvigneshwars: We can just curl the latest tag and use that!

.github/workflows/frontend.yaml
iamvigneshwars: relay compiler doesn't like those apollo directives, removing those doesn't hurt the schema

.github/workflows/frontend.yaml
garryod: I think we want to be building against the current state of the proxy, not the latest tag

.github/workflows/frontend.yaml
```diff
@@ -24,6 +24,26 @@ jobs:
       - name: Install dependencies
         run: yarn --frozen-lockfile
 
+      - name: Get latest Graph Proxy Schema run id
+        run: |
+          GRAPH_WORKFLOW_ID=$(gh api "/repos/DiamondLightSource/workflows/actions/workflows/119673956/runs?status=success" --jq ".workflow_runs[0].id")
```
garryod, line 29: I don't think this guarantees that we get the workflow from this commit, just the latest instance of this workflow

.github/workflows/generate-schema.yaml
```diff
@@ -0,0 +1,34 @@
+name: Generate Supergraph Schema
+
+on:
+  workflow_call:
+
+jobs:
+  generate_schema:
+    # Deduplicate jobs from pull requests and branch pushes within the same repo.
+    if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name != github.repository
```
garryod, line 9: This de-duplication assumes we run on `pull_request` and `push`, so it should be included in the calling workflows instead of here

.github/workflows/generate-schema.yaml
```diff
@@ -0,0 +1,34 @@
...
+          cargo run
+          schema
+          --path workflows.graphql
+        working-directory: graph-proxy
+
+      - name: Upload Schema Artifact
+        uses: actions/upload-artifact@v4.4.0
+        with:
+          name: graphql-schema
+          path: ./graph-proxy/workflows.graphql
```
garryod, line 34: I think this should be taken out and only present in the "Graph Proxy Schema" workflow

.github/workflows/frontend.yaml
```diff
@@ -5,9 +5,70 @@ on:
...
 jobs:
+  generate_schema:
+    uses: ./.github/workflows/generate-schema.yaml
+
+  compose_supergraph:
+    # Deduplicate jobs from pull requests and branch pushes within the same repo.
+    if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name != github.repository
+    runs-on: ubuntu-latest
+    needs: generate_schema
+    steps:
```
garryod, line 16: Can we use a reusable workflow here too? Looks like the commit stuff can all be placed after in the `federation` one

garryod (changes requested):

.github/workflows/generate-schema.yaml
iamvigneshwars: Since we generate and share the schema with `compose_supergaph` job, we need it there!

garryod: Can we not move this `step` into that `job`?

garryod (approved):
LGTM

garryod (approved):
LGTM

PR #217, opened 2024-10-11
Add workflow submission mutation to `graph-proxy`

garryod:
Adds a mutation to the `graph-proxy` service, allowing submission of a workflow using the a `ClusterWorkflowTemplate`.

A few changes were necessary for compatibility:
- `Workflow` `status` is made nullable to handle the case where a workflow has just been submitted and has not yet been given a status by the Argo Workflows Controller - a race condition not previously observed
- Namespacing was removed for template arguments (e.g. `numpy-test.size` -> `size`) to allow these values to be used in GraphQL inputs which accept `[a-zA-Z][_a-zA-Z0-9]*`

Also includes a few other small changes:
- Visit parameters are now captured and returned as a struct for the sake of code re-use
- `async-graphql`'s `Json` type is used in place of custom scalars for `parameters` and `uiSchema`

**Example Query:**

```graphql
mutation {
  submitWorkflowTemplate(
    name: "numpy-benchmark",
    visit: {
      proposalCode: "mg",
      proposalNumber: 36964,
      number: 1
    },
    parameters: {
    	memory: "10Gi",
      size: 1000
    }
  ) {
    name,
    visit {
      proposalCode,
      proposalNumber,
      number
    },
    status {
      __typename
    }
  }
}
```

**Example Response:**

```json
{
  "data": {
    "submitWorkflowTemplate": {
      "name": "numpy-benchmark-8fbfn",
      "visit": {
        "proposalCode": "mg",
        "proposalNumber": 36964,
        "number": 1
      },
      "status": null
    }
  }
}
```

### DIFF
--- a/docs/explanations/parameter-schema-annotations.md
+++ b/docs/explanations/parameter-schema-annotations.md
@@ -4,6 +4,6 @@
 
     See also [UI Schema Annotations](./ui-schema-annotation.md)
 
-On `WorkflowTemplate`s (`workflowtemplates.argoproj.io`) and `ClusterWorkflowTemplate`s (`clusterworkflowtemplates.argoproj.io`) annotations prefixed with `workflows.diamond.ac.uk/paramter-schema` annotations with dot seperated subpaths of the step to which the parameter belongs and the parameter name (e.g. `metadata.annotations."workflows.diamond.ac.uk/parameter-schema.numpy-test.num-cores"`) are reserved for the specification of parameter [JSON Schema](https://json-schema.org/).
+On `WorkflowTemplate`s (`workflowtemplates.argoproj.io`) and `ClusterWorkflowTemplate`s (`clusterworkflowtemplates.argoproj.io`) annotations prefixed with `workflows.diamond.ac.uk/paramter-schema` annotations with a dot separated of the parameter name (e.g. `metadata.annotations."workflows.diamond.ac.uk/parameter-schema.num-cores"`) are reserved for the specification of a parameter [JSON Schema](https://json-schema.org/).
 
 The Schema is used to enhance type checking in the `WorkflowTemplate` and `ClusterWorkflowTemplate` submission forms. If no Schema is supplied one will be automatically generated in which parameters will be assumed to be of type `String` or `Enum` depending on whether they contain the `enum` declaration.
--- a/graph-proxy/src/graphql/workflow_templates.rs
+++ b/graph-proxy/src/graphql/workflow_templates.rs
@@ -4,7 +4,7 @@ use anyhow::anyhow;
 use argo_workflows_openapi::APIResult;
 use async_graphql::{
     connection::{Connection, CursorType, Edge, EmptyFields, OpaqueCursor},
-    scalar, Context, Object, SimpleObject,
+    Context, Json, Object, SimpleObject,
 };
 use axum_extra::headers::{authorization::Bearer, Authorization};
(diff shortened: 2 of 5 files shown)


docs/explanations/parameter-schema-annotations.md
```diff
@@ -4,6 +4,6 @@
 
     See also [UI Schema Annotations](./ui-schema-annotation.md)
 
-On `WorkflowTemplate`s (`workflowtemplates.argoproj.io`) and `ClusterWorkflowTemplate`s (`clusterworkflowtemplates.argoproj.io`) annotations prefixed with `workflows.diamond.ac.uk/paramter-schema` annotations with dot seperated subpaths of the step to which the parameter belongs and the parameter name (e.g. `metadata.annotations."workflows.diamond.ac.uk/parameter-schema.numpy-test.num-cores"`) are reserved for the specification of parameter [JSON Schema](https://json-schema.org/).
+On `WorkflowTemplate`s (`workflowtemplates.argoproj.io`) and `ClusterWorkflowTemplate`s (`clusterworkflowtemplates.argoproj.io`) annotations prefixed with `workflows.diamond.ac.uk/paramter-schema` annotations with a dot separated of the parameter name (e.g. `metadata.annotations."workflows.diamond.ac.uk/parameter-schema.num-cores"`) are reserved for the specification of parameter [JSON Schema](https://json-schema.org/).
```
daurer, line 7:
```suggestion
On `WorkflowTemplate`s (`workflowtemplates.argoproj.io`) and `ClusterWorkflowTemplate`s (`clusterworkflowtemplates.argoproj.io`) annotations prefixed with `workflows.diamond.ac.uk/paramter-schema` annotations with a dot separated of the parameter name (e.g. `metadata.annotations."workflows.diamond.ac.uk/parameter-schema.num-cores"`) are reserved for the specification of a parameter [JSON Schema](https://json-schema.org/).
```

graph-proxy/src/graphql/workflow_templates.rs
```diff
@@ -246,11 +235,11 @@ impl TryFrom<argo_workflows_openapi::IoArgoprojWorkflowV1alpha1ClusterWorkflowTe
...
                 .annotations
                 .remove("workflows.argoproj.io/description"),
-            arguments: ArgumentSchema::new(
+            arguments: Json(ArgumentSchema::new(
                 workflow_template.spec,
                 &mut workflow_template.metadata.annotations,
-            )?,
-            ui_schema: UiSchema::new(&mut workflow_template.metadata.annotations)?,
+            )?),
+            ui_schema: UiSchema::new(&mut workflow_template.metadata.annotations)?.map(Json),
```
daurer, line 242: just out of curiosity: why are you using `Json()` in the first case and `().map(Json)` in the second?

daurer (approved):
Looks good as far as I can tell. Just a small typo and a curiosity question...

graph-proxy/src/graphql/workflow_templates.rs
garryod: In the first instance `ArgumentSchema::new` returns `ArgumentSchema` so can be simply wrapped to give `Json<ArgumentSchema>`, in the second instance `UiSchema::new` returns `Option<UiSchema>` hence we use `::map` to apply the method to the inner value giving `Option<Json<UiSchema>>` (wrapping would give `Json<Option<UiSchema>>` which we don't want)

daurer (approved):

PR #227, opened 2024-10-17
Template Submission Components

garryod:
Depends on #233 & #235 

Adds a `TemplateSubmissionForm` component which uses [JSONForms](https://jsonforms.io/) to render a parameter entry form alongside the template name, description and a submit button with visit selection


daurer (approved):
LGTM

PR #234, opened 2024-10-22
Increase argo workflow chainsaw test sleep duration to 20s

garryod:
Still not a great fix, but hoping this might stop us seeing timeouts


daurer (approved):
Sure, why not increase the sleep but does it actually help...

garryod:
> Sure, why not increase the sleep but does it actually help...

Unsure, would be good if we could find a way of awaiting properly

PR #236, opened 2024-10-24
Add sourceType to groups chart

daurer:


charts/groups/templates/applications.yaml
```diff
@@ -14,10 +17,12 @@ spec:
   project: {{ $group.name }}
   source:
     repoURL: {{ .repoURL }}
-    path: {{ .path }}
-    targetRevision: {{ .targetRevision }}
+    path: {{ $path }}
+    targetRevision: {{ $target }}
```
garryod, line 21: Not sure why I suggested `$path` / `$target` when this could be inlined

```suggestion
    path: {{ default "." .path }}
    targetRevision: {{ default "HEAD" .targetRevision }}
```

garryod (commented):
Looks good, think I made a bad suggestion though

garryod (approved):
LGTM

PR #246, opened 2024-10-31
Add CORS to graph-proxy

garryod:


daurer (approved):
Looks good, to my level of understanding Rust (which is not very high)

PR #250, opened 2024-11-01
Tidy `sessionspaces` chart

garryod:
Uses the [bitnami common chart](https://github.com/bitnami/charts/tree/main/bitnami/common) to simplify things, keeping it consistent with `graph-proxy` etc.


daurer (commented):
Looks good to me, happy to approve once we know that it deploys correctly on the workflows cluster

garryod:
> Looks good to me, happy to approve once we know that it deploys correctly on the workflows cluster

Chart is deployed and running now

daurer (approved):

PR #263, opened 2024-11-06
Some fixes on the React components

daurer:
basically things I am figuring out as I am trying to use the components in a standalone app.


garryod (approved):
LGTM, some embarrassing oversights there by me :sweat_smile:

PR #264, opened 2024-11-06
Fix `$ref`s when merging parameter schema

garryod:
Parameter schema are now treated as root schema, and their references (`$refs`) are now mutated to point to their own `$defs`.


iamvigneshwars (approved):
LGTM

PR #279, opened 2024-11-19
Skip authentication for artifact-related routes

yousefmoazzam:
This allows requests to the artifact-related routes in the argo workflows REST API to not be redirected to the keycloak login page, similar to how requests to the other argo workflows endpoints are handled.


yousefmoazzam:
I didn't try to be clever with the regex where all the artifact-related endpoints start with either `/artifact*/` or `/input-*/`, since I wasn't overly confident with regexes and with this stuff I never know if the trailing slash is important. Let me know if it's preferable to not just enumerate all the endpoints and be a bit smarter about it.

garryod (approved):
LGTM

PR #282, opened 2024-11-19
Create ingress for `workflows-cluster` Kubernetes API

garryod:
Depends on #283 & https://github.com/loft-sh/vcluster/pull/2289

Exposes the `workflows-cluster` kubernetes API at `kubernetes.workflows.diamond.ac.uk`, allowing connection with the following `kubeconfig`:
```yaml
apiVersion: v1
kind: Config
preferences: {}
current-context: user@workflows
clusters:
  - name: workflows
    cluster:
      server: https://kubernetes.workflows.diamond.ac.uk
contexts:
  - name: user@workflows
    context:
      cluster: workflows
      user: user
users:
  - name: user
    user:
      exec:
        apiVersion: client.authentication.k8s.io/v1beta1
        args:
        - oidc-login
        - get-token
        - --oidc-issuer-url=https://authn.diamond.ac.uk/realms/master
        - --oidc-client-id=workflows-cluster
        - --grant-type=password
        - --oidc-use-access-token
        command: kubectl
        env: null
        interactiveMode: IfAvailable
        provideClusterInfo: false
```


charts/workflows-cluster/values.yaml
```diff
@@ -45,7 +45,7 @@ vcluster:
         addVolumes:
           - name: authn-config
             configMap:
-              name: "workflows-cluster-authn-config"  # TODO: Replace with templated value "{{ .Release.Name }}-authn-config"
+              name: "workflows-cluster-authn-config"  # TODO: Replace with templated value "{{ include "common.names.fullname" $ }}-authn-config"
```
daurer, line 48: any reason why we can't use the templated valued?

daurer (approved):

PR #286, opened 2024-11-20
Expose `argo-workflows` server at `argo-workflows.workflows.diamond.ac.uk`

garryod:
Needs #284 

The first step of migrating to the new subdomain, to be followed up by deployment of the dashboard at `workflows.diamond.ac.uk` with removal of the host from this ingress


daurer (approved):
LGTM

PR #288, opened 2024-11-21
Add artifact access via graph proxy

yousefmoazzam:


graph-proxy/src/graphql/workflows.rs
```diff
@@ -52,13 +64,18 @@ impl Workflow {
     pub(super) fn new(
         value: IoArgoprojWorkflowV1alpha1Workflow,
         visit: Visit,
+        artifacts: Vec<Artifact>,
```
garryod, line 67: Since `artifacts` are derived from the `value.status`, we should compute them in this constructor instead of the resolver - as is done with the `WorkflowStatus` struct

yousefmoazzam: Done the first bullet point in c79eb68

yousefmoazzam:
I'm not overly pleased about passing the argo workflows server URL from `WorkflowsQuery::workflows` all the way down to `Task::new`, it feels clumsy, but I didn't see a way to get the server URL to the `Task` constructor otherwise.

If there's a better way to do that then I'm happy to change this. I have a sneaking suspicion that the rejigging of the design as mentioned in the thread above may be better done sooner rather than later though, to avoid having to deal with this.

yousefmoazzam:
If it helps checking that this works, I've been using the following query to test things:
```gql
{
  workflows(visit: {proposalCode: "mg", proposalNumber: 36964, number: 1}){
    nodes {
      name
      status {
        ...on WorkflowSucceededStatus {
          tasks {
            name,
            artifacts {
              name
              url
            }
          }
        }
      }
    }
  }
}
```

graph-proxy/src/graphql/workflows.rs
```diff
@@ -228,6 +234,38 @@ impl TryFrom<String> for TaskStatus {
...
+        visit: &str,
+        name: &str,
+        mut server_url: Url,
+    ) -> Artifact {
+        let artifact_name = &artifact.name[..];
+        server_url.path_segments_mut().unwrap().extend([
+            "artifacts",
+            visit,
+            name,
+            name,
```
garryod, line 259: Surely this should be the `nodeId`? Which presumably comes from the task?

graph-proxy/src/graphql/workflows.rs
```diff
@@ -131,13 +135,14 @@ impl WorkflowRunningStatus {
     fn new(
         value: IoArgoprojWorkflowV1alpha1WorkflowStatus,
         metadata: Arc<Metadata>,
+        server_url: Url,
```
garryod, line 138: I don't love passing the `server_url` in here, we should retrieve it from the `Context` of the `url` resolver

garryod (changes requested):

graph-proxy/src/graphql/workflows.rs
garryod: I'd rather we did the reorganisation now rather than being in this in-between state, happy to sit down an peer-program it if you'd fancy that

garryod:
> I'm not overly pleased about passing the argo workflows server URL from `WorkflowsQuery::workflows` all the way down to `Task::new`, it feels clumsy, but I didn't see a way to get the server URL to the `Task` constructor otherwise.
> 
> If there's a better way to do that then I'm happy to change this. I have a sneaking suspicion that the rejigging of the design as mentioned in the thread above may be better done sooner rather than later though, to avoid having to deal with this.

Another option would be to have the `url` field use a resolver, and store the additional information needed to construct the path in the `Artifact`

graph-proxy/src/graphql/workflows.rs
yousefmoazzam: Yeah that's right, one of the `name`'s is being used as the `nodeId` part in the endpoint, which is:
```
/artifacts/{namespace}/{name}/{nodeId}/{artifactName}
```

I noticed that the `nodeId` was always the same as the workflow name, so I just reused the workflow name for both parts in the path.

This was mainly because `IoArgoprojWorkflowV1alpha1Artifact` doesn't contain the `nodeId`, which is instead in `IoArgoprojWorkflowV1alpha1NodeStatus`, so I opted to be lazy and resuse in the `name` parameter.

Could grab the `nodeId` from the node status object and pass that through in addition to `name`, to make it less confusing (I do agree that the double use of `name` is confusing - and maybe it's incorrect to assume that the workflow name is always the same as the node ID).

garryod: I would only do it this way if they were guaranteed to be the same, but somehow I doubt it's the case if they're exposed as different values

graph-proxy/src/graphql/workflows.rs
yousefmoazzam: There's a separate PR for the refactoring https://github.com/DiamondLightSource/workflows/pull/376, so we can move any further discussion to there.

graph-proxy/src/graphql/workflows.rs
yousefmoazzam: The appropriate values are now used to form the endpoint, so hopefully this issue is resolved now.

There's still the question of if what I do now with passing individual values from the workflow manifest to construct an `Artifact` is acceptable, or if the entire manifest (wrapped in an `Arc`) should be passed along and stored in an `Artifact`, but that's a slightly different topic and can be discussed separately.

yousefmoazzam:
I kept the second commit separate because, while I'm fairly sure the filename is what we want to give requestors rather than the artifact name which argo workflows uses, maybe we'd want to provide the artifact name, so I left the commits separate. Can squash if the filename is what we want.

Also, lots of assumptions being made in getting the filename:
1. argo workflow artifacts always stored in an s3 bucket (probably true, config seems to indicate this, but maybe a better way to handle this)
2. the "key" of an artifact should be non-null (seems reasonable, but I don't know much about s3 buckets, I went off of the fact that a "key" looks like a filepath relative to a bucket)
3. an artifact "key" shouldn't have `..` at the end of it
4. filename should be valid utf-8

graph-proxy/src/graphql/workflows.rs
```diff
@@ -275,16 +275,60 @@ impl TryFrom<String> for TaskStatus {
     }
 }
 
+/// A task artifact
```
garryod, line 278: This feels like it could be a bit more descriptive, what is an artifact?

graph-proxy/src/graphql/workflows.rs
```diff
@@ -275,16 +275,60 @@ impl TryFrom<String> for TaskStatus {
...
+#[derive(Debug)]
+struct Artifact<'a> {
+    manifest: &'a IoArgoprojWorkflowV1alpha1Artifact,
+    metadata: &'a Metadata,
+    node_id: &'a str,
+}
+
+#[Object]
+impl Artifact<'_> {
+    /// The name of the artifact
```
garryod, line 289: This should probably say we're using the file name

graph-proxy/src/graphql/workflows.rs
```diff
@@ -275,16 +275,60 @@ impl TryFrom<String> for TaskStatus {
...
+    async fn name(&self) -> &str {
+        let path = std::path::Path::new(
+            self.manifest
+                .s3
+                .as_ref()
+                .expect("Artifacts are stored in an S3 bucket")
```
garryod, line 295: Instead of panicking, could we return some error variant (e.g. `UnrecognisedArtifactStore`)?

graph-proxy/src/graphql/workflows.rs
```diff
@@ -275,16 +275,60 @@ impl TryFrom<String> for TaskStatus {
...
+                .key
+                .as_ref()
+                .expect("Stored artifact should have non-null key"),
```
garryod, line 298: Instead of panicking, could we return an error variant?

graph-proxy/src/graphql/workflows.rs
```diff
@@ -275,16 +275,60 @@ impl TryFrom<String> for TaskStatus {
...
+    manifest: &'a IoArgoprojWorkflowV1alpha1Artifact,
+    metadata: &'a Metadata,
+    node_id: &'a str,
+}
+
+#[Object]
+impl Artifact<'_> {
+    /// The name of the artifact
+    async fn name(&self) -> &str {
+        let path = std::path::Path::new(
```
garryod, line 291: You can import `Path` :sweat_smile:

graph-proxy/src/graphql/workflows.rs
```diff
@@ -275,16 +275,60 @@ impl TryFrom<String> for TaskStatus {
...
+            self.manifest
+                .s3
+                .as_ref()
+                .expect("Artifacts are stored in an S3 bucket")
+                .key
+                .as_ref()
+                .expect("Stored artifact should have non-null key"),
+        );
+        path.file_name()
+            .expect("Path to artifact shouldn't end with `..`")
```
garryod, line 301: Would it perhaps make more sense to use `ok_or_default` here? And just return an empty string (i.e. `""`) if it's somehow `None`

graph-proxy/src/graphql/workflows.rs
```diff
@@ -275,16 +275,60 @@ impl TryFrom<String> for TaskStatus {
...
+            .to_str()
+            .expect("Filename should be valid utf-8 encoded string")
```
garryod, line 303: Instead of panicking, could we return an error variant?

graph-proxy/src/graphql/workflows.rs
```diff
@@ -318,6 +362,22 @@ impl Task {
     async fn dependencies(&self) -> Vec<String> {
         self.node_status.children.clone()
     }
+
+    /// Artifacts produced by a task
+    async fn artifacts(&self) -> Vec<Artifact> {
+        match self.node_status.outputs.as_ref() {
+            None => vec![],
```
garryod, line 369: This it a nit-pick, but we usually use `Vec::new` for empty Vecs

garryod (changes requested):
Looking good, could make more use of error types though, just don't want to crash in prod

graph-proxy/src/graphql/workflows.rs
yousefmoazzam: No worries, I wasn't aware of the convention, that's useful to know

graph-proxy/src/graphql/workflows.rs
```diff
@@ -318,6 +368,22 @@ impl Task {
...
+            Some(outputs) => outputs
+                .artifacts
+                .iter()
+                .map(|manifest| Artifact {
+                    manifest,
+                    metadata: self.metadata,
+                    node_id: &self.node_status.id,
+                })
+                .collect::<Vec<_>>(),
+        }
```
garryod, line 385: I might be tempted to `map` here

```suggestion
        self.node_status.outputs.as_ref().map(|outputs| {
            outputs
                .artifacts
                .iter()
                .map(|manifest| Artifact {
                    manifest,
                    metadata: self.metadata,
                    node_id: &self.node_status.id,
                })
                .collect::<Vec<_>>()
        }).unwrap_or_default()
```

garryod (approved):
LGTM

graph-proxy/src/graphql/workflows.rs
yousefmoazzam: I like it, done!

garryod (approved):
LGTM

PR #289, opened 2024-11-21
Build dashboard with nginx container

daurer:
- [x] Add Dockerfile for building dashboard app with nginx
- [x] Add CI step for building and publishing dashboard container

### DIFF
--- a/frontend/Dockerfile
+++ b/frontend/Dockerfile
@@ -1,0 +1,25 @@
+FROM docker.io/library/node:20-bookworm-slim AS build
+
+WORKDIR /app
+
+COPY ./yarn.lock ./package.json .
+COPY ./dashboard/package.json ./dashboard/package.json
+COPY ./relay-workflows-lib ./relay-workflows-lib
+COPY ./workflows-lib ./workflows-lib
+RUN yarn install --frozen-lockfile
+
+COPY . .
+COPY ./supergraph.graphql ./relay-workflows-lib/supergraph.graphql
+RUN yarn run --cwd=relay-workflows-lib relay
+
+COPY ./supergraph.graphql ./dashboard/supergraph.graphql
+RUN yarn run --cwd=dashboard build
+
+FROM docker.io/library/nginx:1.23.3-alpine-slim as nginx
+
+WORKDIR /app
+
+COPY --from=build /app/dashboard/dist /usr/share/nginx/html
+COPY ./container_utils/default.conf.template /etc/nginx/templates/default.conf.template
+
+EXPOSE 80
--- a/frontend/dashboard/relay.config.json
+++ b/frontend/dashboard/relay.config.json
@@ -1,7 +1,7 @@
 {
   "src": "./src",
   "language": "typescript",
-  "schema": "./src/workflows.graphql",
+  "schema": "./supergraph.graphql",
   "exclude": [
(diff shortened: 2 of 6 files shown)


frontend/dashboard/relay.config.json
```diff
@@ -1,7 +1,7 @@
 {
   "src": "./src",
   "language": "typescript",
-  "schema": "./src/workflows.graphql",
+  "schema": "../relay-workflows-lib/supergraph.graphql",
```
garryod, line 4: This feels a little concerning, is there a better way of passing the schema at build time? If not perhaps we should lift the `supergraph.graphql` file up into the `frontend` dir?

frontend/Dockerfile
```diff
@@ -0,0 +1,26 @@
+FROM docker.io/library/node:20-bookworm-slim AS build
+
+WORKDIR /app
+
+COPY ./yarn.lock /app/yarn.lock
+COPY ./package.json /app/package.json
```
garryod, line 6: Since we're copying to the `WORKDIR` we can use the shorthand version
```suggestion
COPY ./yarn.lock package.json .
```

frontend/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+COPY ./dashboard/package.json /app/dashboard/package.json
+COPY ./relay-workflows-lib /app/relay-workflows-lib
+COPY ./workflows-lib /app/workflows-lib
```
garryod, line 9: We should probably use `./` in place of `/app/` here

frontend/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+RUN yarn install --frozen-lockfile
+
+COPY . .
+WORKDIR /app/relay-workflows-lib
+RUN yarn run relay
```
garryod, line 14: Typically we don't want to change the `WORKDIR` frequently
```suggestion
RUN yarn run --cwd relay-workflows-lib relay
```

frontend/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+
+WORKDIR /app/dashboard
+RUN yarn run relay && yarn tsc && yarn build --outDir ./static
```
garryod, line 17: These should probably be condensed into one yarn command, in the `package.json` `scripts`

garryod (changes requested):

frontend/dashboard/relay.config.json
daurer: Yeah, I was not so sure about this. I think we could either

- copy the schema into multiple subdirectories in the Dockerfile
- move the schema into the frontend dir and point all the relay configs there

frontend/Dockerfile
daurer: Oh, I was not aware of the `--cwd` flag in yarn. That is much better.

frontend/Dockerfile
daurer: The existing `build` command from the scripts already does the job. Also changed to using the default `dist` output directory for the static build

frontend/Dockerfile
daurer: Need to use `--cwd=relay-workflows-lib` to get this to work

garryod (approved):

PR #290, opened 2024-11-22
Add helm chart for workflows dashboard

daurer:
- [x] Added new helm chart for the workflows dashboard -> now available at workflows.diamond.ac.uk
- [x] Removed workflows.diamond.ac.uk from the workflows chart

### DIFF
--- a/charts/argocd-apps/Chart.yaml
+++ b/charts/argocd-apps/Chart.yaml
@@ -3,4 +3,4 @@ name: argocd-apps
 description: An argocd app to deploy apps inside the virtual cluster
 type: application
 
-version: 0.3.2
+version: 0.4.0
--- a/charts/argocd-apps/dev-values.yaml
+++ b/charts/argocd-apps/dev-values.yaml
@@ -52,3 +52,6 @@ workflows:
 
 graphProxy:
   enabled: true
+
+dashboard:
+  enabled: true
--- a/charts/dashboard/Chart.yaml
+++ b/charts/dashboard/Chart.yaml
@@ -1,0 +1,12 @@
+apiVersion: v2
+name: dashboard
+description: A dashboard for Diamond workflows
+type: application
+
+version: 0.1.0
+appVersion: 0.1.0-rc23
+
+dependencies:
+  - name: common
+    version: 2.23.0
+    repository: oci://docker.io/bitnamicharts
--- a/charts/dashboard/templates/deployment.yaml
+++ b/charts/dashboard/templates/deployment.yaml
@@ -1,0 +1,71 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ include "common.names.fullname" $ }}
+  namespace: {{ .Release.Namespace }}
+  {{- with $.Values.deployment.annotations }}
+  annotations:
+    {{- . | toYaml | nindent 4 }}
+  {{- end }}
+  labels:
+    {{- include "common.labels.standard" $ | nindent 4 }}
+spec:
+  replicas: {{ $.Values.deployment.replicas }}
+  selector:
+    matchLabels:
+      {{- include "common.labels.matchLabels" $ | nindent 6 }}
+  template:
+    metadata:
+      {{- with $.Values.deployment.podAnnotations }}
+      annotations:
+        {{- . | toYaml | nindent 8 }}
+      {{- end }}
+      labels:
+        {{- include "common.labels.matchLabels" $ | nindent 8 }}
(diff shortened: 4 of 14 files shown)


charts/argocd-apps/Chart.yaml
```diff
@@ -3,4 +3,4 @@ name: argocd-apps
 description: An argocd app to deploy apps inside the virtual cluster
 type: application
 
-version: 0.3.2
+version: 0.3.3
```
garryod, line 6: This should be minor as it adds a whole app

charts/argocd-apps/dev-values.yaml
```diff
@@ -52,3 +52,6 @@ workflows:
 
 graphProxy:
   enabled: true
+
+dashboard:
+  enabled: false
```
garryod, line 57: Do we not want this enabled?

charts/dashboard/Chart.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: v2
+name: dashboard
+description: A dashboard for Diamond workflows
+type: application
+
+version: 0.1.0
+appVersion: 0.1.0-rc1
```
garryod: Probably want to use the most recent tag on the repo here

charts/dashboard/templates/deployment.yaml
```diff
@@ -0,0 +1,55 @@
...
+      {{- else }}
+      serviceAccountName: {{ default "default" $.Values.serviceAccount.name }}
+      {{- end }}
+      imagePullSecrets: {{ include "common.images.renderPullSecrets" (dict "images" (list $.Values.image) "context" $ ) }}
+      containers:
+        - name: dashboard
+          image: {{ include "common.images.image" ( dict "imageRoot" $.Values.image "global" $.Values.global "chart" $.Chart ) }}
+          imagePullPolicy: {{ $.Values.image.pullPolicy }}
+          args:
+            - serve
```
garryod, line 37: This seems wrong

charts/dashboard/templates/deployment.yaml
garryod, line 1: Probably want to set up a health and liveness check since `nginx` is configured to provide a healthz endpoint

charts/dashboard/values.yaml
garryod, line 1: This is missing the `service` fields

charts/argocd-apps/dev-values.yaml
daurer: I thought it might not work anyway to look at the dashboard if the namespace is not `workflows` that's why I disabled it, but happy to enable

charts/dashboard/Chart.yaml
daurer: Yeah, saw that one myself and already changed it to rc-23

charts/dashboard/templates/deployment.yaml
daurer: Removed

charts/dashboard/values.yaml
daurer: Added service fields and `service.yaml` in templates

charts/argocd-apps/Chart.yaml
daurer: changed to 0.4.0

charts/argocd-apps/dev-values.yaml
daurer: Okay, will enable it then

charts/dashboard/templates/deployment.yaml
daurer: I have added liveness, readiness and startup probes pointing to /healthcheck as discussed offline

charts/dashboard/templates/deployment.yaml
```diff
@@ -0,0 +1,59 @@
...
+          {{- with $.Values.deployment.tolerations }}
+          tolerations:
+            {{- . | toYaml | nindent 8 }}
+          {{- end }}
+          livenessProbe:
+            {{- $.Values.deployment.livenessProbe | toYaml | nindent 12 }}
+          readinessProbe:
+            {{- $.Values.deployment.readinessProbe | toYaml | nindent 12 }}
+          startupProbe:
+            {{- $.Values.deployment.startupProbe | toYaml | nindent 12 }}
```
garryod, line 59: I think we can afford to hard code these since we know the nature of the container being run

garryod (changes requested):
Looks good, just one comment on the probes

garryod (approved):
LGTM

PR #297, opened 2024-11-26
Bump charts to latest app version

daurer:
Updated sessionspaces, graph-proxy and dashboard charts to use  `0.1.0-rc24`


garryod (approved):
LGTM

PR #302, opened 2024-11-27
Fix dashboard CORS error by using correctly configured client

garryod:
Uses the `workflows-dashboard` client, which is configured with the correct Web Origins (i.e. `Access-Control-Allow-Origin`)


daurer (approved):
LGTM

PR #303, opened 2024-11-27
add info logs to graph-proxy main

iamvigneshwars:


graph-proxy/src/main.rs
```diff
@@ -110,11 +114,18 @@ async fn main() {
...
     schema: RootSchema,
     prefix_path: &str,
     cors_allow: Option<Vec<Regex>>,
 ) -> anyhow::Result<Router> {
+    info!("Setting up the router");
+    if let Some(cors_allow) = &cors_allow {
+        info!("CORS rules enabled: {:?}", cors_allow);
+    } else {
+        info!("CORS rules disabled. Allowing default origin.");
+    }
```
garryod, line 128: Could we move these into the `if` / `else` statement below?

garryod (approved):
LGTM, cheers

PR #305, opened 2024-11-28
Fix `graph-proxy` docker image caching

garryod:
By only copying the necessary build files (`Cargo.toml`, `build.rs` & `lib.rs`) we can ensure cache hits for the graph-proxy build, thus significantly speeding up build times for the container in CI
By ensuring the temporary `main.rs` has a consistent time stamp the layer cache for the initial `cargo build --release` is not invalidated

Previously: `5m17s`
Now: `29s`


daurer (approved):
LGTM, wow that's an impressive improvement in speed!

PR #308, opened 2024-11-29
feat: release please setup to automate releases

iamvigneshwars:


garryod (changes requested):
A few thoughts:
- How do we want to handle the helm linting which at present requires us to bump the chart version with each change - presumably this should only now be bumped by release-please?
- This will still publish containers for every tag, which we don't want
- Is there a a reason we separate the docs? And what if the `mkdocs.yaml` is updated?
- How do we ensure that commits are conventional? As non-conventional ones will break our usage of release-please

.github/workflows/_graph_proxy_container.yaml
```diff
@@ -42,8 +42,8 @@ jobs:
           build-args: ARGO_SERVER_SCHEMA_URL=https://raw.githubusercontent.com/argoproj/argo-workflows/main/api/jsonschema/schema.json
           context: ./graph-proxy
           target: deploy
-          push: ${{ github.event_name == 'push' && startsWith(github.ref, 'refs/tags') }}
-          load: ${{ !(github.event_name == 'push' && startsWith(github.ref, 'refs/tags')) }}
+          push: ${{ github.event_name == 'release' && startsWith(github.event.release.tag_name, 'graph-proxy@') }}
+          load: ${{ !(github.event_name == 'release' && startsWith(github.event.release.tag_name, 'graph-proxy@')) }}
```
garryod, line 46: Does this ever actually get called with a `release` event?

garryod (changes requested):

release-please-config.json
```diff
@@ -0,0 +1,61 @@
...
+    },
+    "charts/verflixt": {
+      "release-type": "helm"
+    },
+    "charts/workflows": {
+      "release-type": "helm"
+    },
+    "charts/workflows-cluster": {
+      "release-type": "helm"
+    },
```
garryod, line 47: We probably don't need to release the helm charts as our ArgoCD instance points to `main`

release-please-config.json
```diff
@@ -0,0 +1,61 @@
...
+  "tag-separator": "@",
+  "separate-pull-requests": true,
+  "extra-files": [
+    "README.md",
+    "mkdocs.yml"
+  ],
+  "packages": {
+    "backend/graph-proxy": {
+      "release-type": "rust"
+    },
```
garryod, line 14: We should update the `appVersion` in the corresponding charts `Chart.yaml` when we release a new container, this can be [done with extra-files](https://github.com/googleapis/release-please/blob/main/docs/customizing.md#updating-arbitrary-yaml-files)

garryod (changes requested):

release-please-config.json
daurer: removed all charts and also docs from automated releasing

release-please-config.json
daurer: done

daurer:
Have tested in a fork that 

- [x] PRs are created as expected including bump in appVersion of Chart.yaml
- [x] After merging the PR,  tags are created, e.g. <service>@v0.1.0
- [x] Also, containers are created and pushed with expected tags, e.g. ghcr.io/.../<service>:0.1.0

Things to improve in the future

- `node-workspace` and `cargo-workspace` to also create releases when dependencies are bumped in `/frontend` or `backend`

.github/workflows/_release_please.yaml
```diff
@@ -0,0 +1,24 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Create GitHub App Token
+        id: app-token
+        uses: actions/create-github-app-token@v1.11.0
+        with:
+          app-id: 1077859
+          private-key: ${{ secrets.release_please_token }}
+
+      - uses: googleapis/release-please-action@v4
```
garryod, line 20: Can we use a fully qualified version here?
```suggestion
      - uses: googleapis/release-please-action@v4.1.3
```

release-please-config.json
garryod, line 1: It would probably be good to include a `bootstrap-sha` in here, just to mark where we're beginning release-please. Can always manually create releases if we want to publish something that hasn't been updated since

release-please-config.json
```diff
@@ -0,0 +1,54 @@
...
+  "include-component-in-tag": true,
+  "include-v-in-tag": true,
+  "tag-separator": "@",
+  "separate-pull-requests": true,
+  "bump-minor-pre-major": true,
+  "bump-patch-for-minor-pre-major": true,
+  "extra-files": [
+    "README.md",
+    "mkdocs.yml"
+  ],
```
garryod, line 12: What do these `extra-files` do in this context?

garryod (changes requested):

release-please-config.json
daurer: done

release-please-config.json
daurer: this was hangover from releasing docs

garryod (approved):
LGTM

daurer (approved):

garryod (approved):

PR #323, opened 2024-12-09
Add graph proxy workflow resolver function tests

yousefmoazzam:
This adds tests of the various queries, before the reorganisation of the code is done as mentioned in https://github.com/DiamondLightSource/workflows/pull/288#discussion_r1853857325


yousefmoazzam:
@iamvigneshwars This is a first attempt at testing the workflow resolver functions, would be interesting to hear if you have any thoughts on this

iamvigneshwars:
Is there any reason for not using the well maintained `mockito` crate 🤔

yousefmoazzam:
No reason at all, the first hits I got when searching something like "http mock rust" were `httpmock` and `wiremock`, so I just went with `httpmock` because it looked simple enough to use.

I can see `mockito` indeed has more stars than either, I'll try it out and see how it compares; cheers for pointing out `mockito`!

.github/workflows/_graph_proxy_code.yaml
```diff
@@ -63,5 +63,6 @@ jobs:
         working-directory: graph-proxy
         run: >
           cargo test
+          --release
```
garryod, line 66: Why do we need `--release` here?

graph-proxy/src/graphql/workflows.rs
```diff
@@ -445,3 +445,419 @@ impl WorkflowsQuery {
...
+        let query = format!(
+            r#"
+            query {{
+                workflow(name: "{}", visit: {{proposalCode: "{}", proposalNumber: {}, number: {}}}) {{
+                    name
+                }}
+            }}
+        "#,
+            workflow_name, proposal_code, proposal_number, number
+        );
```
garryod, line 495: Is there a better way to build this? Should we be using a GraphQL client for this?

graph-proxy/src/graphql/workflows.rs
```diff
@@ -445,3 +445,419 @@ impl WorkflowsQuery {
...
+                .header("content-type", "application/json")
+                .body(std::fs::read_to_string(response_file_path).unwrap());
+        });
+
+        let token = None::<Authorization<Bearer>>;
+        let argo_server_url = Url::parse(&server.base_url()).unwrap();
+        let schema = root_schema_builder()
+            .data(ArgoServerUrl(argo_server_url))
+            .data(token)
+            .finish();
```
garryod, line 485: Can inline the `token` here as it's not used anywhere else

```suggestion
        let argo_server_url = Url::parse(&server.base_url()).unwrap();
        let schema = root_schema_builder()
            .data(ArgoServerUrl(argo_server_url))
            .data(None::<Authorization<Bearer>>)
            .finish();
```

garryod (changes requested):

graph-proxy/src/graphql/workflows.rs
```diff
@@ -445,3 +445,419 @@ impl WorkflowsQuery {
...
+        workflow_endpoint.assert();
+        let mut outer_map = IndexMap::new();
+        let outer_map_key = Name::new("workflow");
+        let mut inner_map = IndexMap::new();
+        let inner_map_key = Name::new("name");
+        let inner_map_value = Value::from(workflow_name);
+        inner_map.insert(inner_map_key, inner_map_value);
+        outer_map.insert(outer_map_key, Value::from(inner_map));
+        let expected_data = Value::from(outer_map);
+        assert_eq!(resp.data, expected_data);
```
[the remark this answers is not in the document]
yousefmoazzam: Thanks for the suggestion, that would make building the expected response much more readable, will do

.github/workflows/_graph_proxy_code.yaml
yousefmoazzam: I found that running certain tests that make graphql queries with `cargo test` (no  `--release`) suffers the same issue as when running the graph-proxy with `cargo run serve` (no  `--release`): tokio reports a stack overflow when handling a query, and the test fails as a consequence.

I'll take another look at that stack overflow issue; revisiting the github issue I looked at before which mentioned the same issue, it looks like it was resolved early in December https://github.com/async-graphql/async-graphql/issues/1506

graph-proxy/src/graphql/workflows.rs
yousefmoazzam: Yeah you're probably right, this way is a bit fiddly, with the need to escape the curly braces - I'll try out using a proper client instead.

graph-proxy/src/graphql/workflows.rs
yousefmoazzam: Ah good point, will make that change in all the tests

graph-proxy/src/graphql/workflows.rs
yousefmoazzam: Eh, after looking into some graphql clients, it feels like some are a bit overkill:
- [`graphql-client`](https://github.com/graphql-rust/graphql-client?tab=readme-ov-file#getting-started) is the most popular rust client, but it needs a file with the schema, and a file with the desired query, which feels unnecessary for the purposes of testing using small queries
- similar thing for [`cynic`](https://cynic-rs.dev/quickstart#fetching-the-graphql-schema), it needs a copy of the schema in a file
- I found [`reqwest-graphql`](https://docs.rs/reqwest-graphql/latest/reqwest_graphql/#basic-usage) which doesn't need the schema in a file
  - it uses queries in raw strings, which is similar to what is currently being done
  - though it does allow for the use of stuff like `$` in the raw string
  - and unfortunately, this crate doesn't look well-maintained

I'm leaning towards keeping the query building as-is and not using a client, let me know what you think.

garryod: Yeah, none of those look like terrific options, a `format!` string seems like the call

graph-proxy/src/graphql/workflows.rs
```diff
@@ -445,3 +445,384 @@ impl WorkflowsQuery {
...
+        workflows_endpoint.assert();
+        let nodes = ["numpy-benchmark-wdkwj", "numpy-benchmark-n6jsg"]
+            .iter()
+            .map(|name| json!({"name": name}))
+            .collect::<Vec<_>>();
+        let expected_data = json!({
+            "workflows": {
+                "nodes": nodes
+            }
+        });
```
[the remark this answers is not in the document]
yousefmoazzam: No worries; if the objects had multiple fields then maybe `map` would be worth it, but I agree for this case it's unnecessary, and clearer to just write the objects explicitly

yousefmoazzam:
Finally got it working with `mockito`, I think I prefer it for two reasons (in addition to being better maintained as pointed out by @iamvigneshwars):
- it caught an issue which wasn't caught by `httpmock` (which is why I added another test in ee31e69)
- the order of the definition of the two mock endpoints in the pagination test doesn't matter with `mockito`, but it did matter with `httpmock` (one ordering failed, the other passed)

.github/workflows/_graph_proxy_code.yaml
yousefmoazzam: Ok, I used a `.cargo/config.toml` file to specify the `RUST_MIN_STACK` env var. [Docs](https://doc.rust-lang.org/cargo/reference/config.html#configuration-format) say this will be present for any process that `cargo` runs, so this also gets rid of the stack overflow when running `cargo run serve`.

I'm not familiar with this `.cargo/config.toml` file though, so let me know if this isn't quite the right config.

(The other thing I found in rust docs for setting env vars was [build script stuff](https://doc.rust-lang.org/cargo/reference/build-scripts.html#rustc-env), though they explicitly say this should only be used for env var inspection at compile-time using `env!`).

graph-proxy/.cargo/config.toml
```diff
@@ -0,0 +1,2 @@
+[env]
+RUST_MIN_STACK = "4194304"
```
garryod, line 2: Would be nice to have a human readable size here

```suggestion
RUST_MIN_STACK = "4194304" # 4MiB
```

graph-proxy/src/graphql/workflows.rs
```diff
@@ -445,3 +445,457 @@ impl WorkflowsQuery {
...
+
+#[cfg(test)]
+mod tests {
+    use std::path::PathBuf;
+
+    use serde_json::json;
+    use url::Url;
+
+    use super::*;
+    use crate::graphql::root_schema_builder;
```
garryod, line 457: Nit: would rather if these weren't separated, and we didn't use the wildcard import on `super`

graph-proxy/src/graphql/workflows.rs
```diff
@@ -445,3 +445,457 @@ impl WorkflowsQuery {
...
+
+    #[tokio::test]
+    async fn single_workflow_query() {
+        let workflow_name = "numpy-benchmark-wdkwj";
+        let proposal_code = "mg";
+        let proposal_number = 36964;
+        let number = 1;
+        let namespace = format!("{}{}-{}", proposal_code, proposal_number, number);
```
garryod, line 465: We should probably use `crate::graphql::Visit` here. The `to_string` of which would produce your `namespace`

graph-proxy/src/graphql/workflows.rs
yousefmoazzam: Nice, I didn't know `rustfmt` ordered imports for you if there's no newlines between them

garryod (approved):
LGTM

PR #324, opened 2024-12-10
Monitoring Stack setup

iamvigneshwars:


garryod (changes requested):
Looking good, any chance of a standalone app so we can deploy this with the argus ArgoCD?

charts/monitoring/values.yaml
```diff
@@ -0,0 +1,252 @@
...
+
+  ingress:
+    enabled: true
+    path: "/"
+    hosts:
+      - grafana.workflows.diamond.ac.uk
+    tls:
+    - hosts:
+      - grafana.workflows.diamond.ac.uk
+      secretName: letsencrypt-grafana-workflows-diamond-ac-uk
```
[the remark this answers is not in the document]
iamvigneshwars: Think we should just include it as a sealed secret

charts/monitoring/values.yaml
```diff
@@ -0,0 +1,393 @@
...
+      enabled: false
+  queryFrontend:
+    ingress:
+      enabled: false
+    resources:
+      limits:
+        cpu: 250m
+        memory: 2Gi
+    networkPolicy:
+      enabled: false
```
garryod, line 78: Do we need these if it's disabled?

charts/monitoring/values.yaml
```diff
@@ -0,0 +1,393 @@
...
+    enabled: false
+  podSecurityPolicy:
+    enabled: false
+  rbac:
+    create: false
+  alertmanager:
+    serviceAccount:
+      create: false
+      annotations: {}
+      name: default-full-access-mounted
```
garryod, line 145: Can we avoid using `default-full-access-mounted`? Looks like we might need to create our own `Role` & `RoleBinding` here

garryod (changes requested):

charts/monitoring/values.yaml
```diff
@@ -0,0 +1,261 @@
...
+          - grafana.workflows.diamond.ac.uk
+        secretName: letsencrypt-grafana-workflows-diamond-ac-uk
+
+  datasources:
+    datasources.yaml:
+      apiVersion: 1
+      datasources:
+      - name: Prometheus
+        type: prometheus
+        url: "http://{{ .Release.Name }}-thanos-query-frontend:9090"
```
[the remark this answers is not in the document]
iamvigneshwars: I think `.Release.Name` would be more accurate here

charts/monitoring/values.yaml
iamvigneshwars: By default it's enabled

charts/monitoring/values.yaml
iamvigneshwars: seems to work fine without creating `Role` & `RoleBinding`

charts/monitoring/values.yaml
```diff
@@ -0,0 +1,393 @@
+grafana:
+  rbac:
+    create: false
```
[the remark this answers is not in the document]
iamvigneshwars: Yeah, but we don't want grafana to create any other roles!

charts/monitoring/templates/otel-collector-rolebinding.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  namespace: {{ .Release.Namespace }}
+  name: {{ include "common.names.fullname" . }}-otel-collector
+subjects:
+  - kind: ServiceAccount
+    name: "{{ .Release.Name }}-opentelemetry-collector"
+    namespace: workflows
```
garryod, line 9: This should be templated

charts/monitoring/templates/otel-collector-rolebinding.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  namespace: {{ .Release.Namespace }}
+  name: {{ include "common.names.fullname" . }}-otel-collector
+subjects:
+  - kind: ServiceAccount
+    name: "{{ .Release.Name }}-opentelemetry-collector"
```
garryod, line 8: Can we stick with either `otel-collector` or `opentelemetry-collector` throughout?

charts/monitoring/values.yaml
garryod: Ah, my bad, I misread the `queryFrontend.ingress.enabled` as simply `queryFrontend.enabled`

garryod (changes requested):

PR #366, opened 2025-01-17
Argocd workflows apps

iamvigneshwars:
depends on #324

### DIFF
--- a/charts/base-apps/Chart.yaml
+++ b/charts/base-apps/Chart.yaml
@@ -1,0 +1,9 @@
+apiVersion: v2
+name: base-apps
+description: "An Argocd app to deploy a monitoring stack and a virtual workfows-cluster which inturn deploys multiple argocd apps inside the workflows-cluster"
+type: application
+version: 0.1.0
+dependencies:
+  - name: common
+    repository: https://charts.bitnami.com/bitnami
+    version: 2.27.0
--- a/charts/base-apps/Chart.yaml
+++ b/charts/base-apps/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: base-apps
-description: "An Argocd app to deploy a monitoring stack and a virtual workfows-cluster which inturn deploys multiple argocd apps inside the workflows-cluster"
+description: An Argocd app to deploy a monitoring stack and a virtual workfows-cluster which inturn deploys multiple argocd apps inside the workflows-cluster
 type: application
 version: 0.1.0
 dependencies:
--- a/charts/base-apps/templates/monitoring-application.yaml
+++ b/charts/base-apps/templates/monitoring-application.yaml
@@ -1,0 +1,30 @@
+{{- if .Values.monitoring.enabled }}
+apiVersion: argoproj.io/v1alpha1
+kind: Application
+metadata:
+  name: {{ include "common.names.fullname" . }}-monitoring
+  namespace: {{ .Release.Namespace }}
+spec:
+  destination:
+    namespace: {{ .Release.Namespace }}
+    server: {{ .Values.destination.server }}
+  project: {{ .Release.Namespace }}
+  source:
+    repoURL: https://github.com/DiamondLightSource/workflows.git
+    path: charts/monitoring
+    targetRevision: {{ .Values.monitoring.targetRevision }}
+    helm:
+      valueFiles:
+        - values.yaml
+        {{- if .Values.monitoring.extraValueFiles }}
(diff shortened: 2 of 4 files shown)


charts/argocd-workflows-apps/Chart.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v2
+name: argocd-workflows-apps
+description: An argocd app to deploy apps in the namespace
```
garryod, line 3: This description could probably be more descriptive - since there are only two apps we can probably enumerate them

charts/argocd-workflows-apps/templates/monitoring-stack-application.yaml
```diff
@@ -0,0 +1,32 @@
+{{- if .Values.monitoring.enabled }}
+apiVersion: argoproj.io/v1alpha1
+kind: Application
+metadata:
+  name: monitoring
+  namespace: workflows
```
garryod, line 6: Namespace needs templated

charts/argocd-workflows-apps/templates/monitoring-stack-application.yaml
```diff
@@ -0,0 +1,32 @@
+{{- if .Values.monitoring.enabled }}
+apiVersion: argoproj.io/v1alpha1
+kind: Application
+metadata:
+  name: monitoring
```
garryod, line 5: Would be good to template this with the `common.names.fullname` as a prefix

charts/argocd-workflows-apps/templates/monitoring-stack-application.yaml
```diff
@@ -0,0 +1,32 @@
...
+  namespace: workflows
+  annotations:
+    argocd.argoproj.io/sync-wave: "2"
```
garryod, line 8: Why sync wave 2?

charts/argocd-workflows-apps/templates/monitoring-stack-application.yaml
```diff
@@ -0,0 +1,32 @@
...
+spec:
+  destination:
+    namespace: workflows
```
garryod, line 11: This should be templated

charts/argocd-workflows-apps/values.yaml
```diff
@@ -0,0 +1,14 @@
+destination:
+  server: https://api.argus.diamond.ac.uk:6443
```
garryod, line 2: It would probably be better to default to the `kubernetes.default.svc` domain, to make the value cluster portable

charts/argocd-workflows-apps/templates/monitoring-stack-application.yaml
```diff
@@ -0,0 +1,32 @@
...
+    server: {{ .Values.destination.server }}
+  project: workflows
```
garryod, line 13: I believe this can (and should) be templated using the namespace, since we map projects to namespaces in a one-to-one fashion

charts/argocd-workflows-apps/templates/workflows-cluster-application.yaml
garryod, line 1: Same comments apply from the `monitoring` app

charts/argocd-workflows-apps/templates/monitoring-stack-application.yaml
garryod, line 1: Would be good if the file name was consistent with the application name (e.g. `monitoring-application`)

garryod (changes requested):
Looks reasonable, just needs a bit of templating to make it portable. Also not sure what I think of the `argocd-workflows-apps` name, could probably do with making it a bit shorter, maybe `host-apps`? Open to suggestions on that one though :sweat_smile:

charts/argocd-workflows-apps/templates/monitoring-stack-application.yaml
iamvigneshwars: Ahh, forgot to remove it!

charts/base-apps/Chart.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v2
+name: base-apps
+description: "An Argocd app to deploy a monitoring stack and a virtual workfows-cluster which inturn deploys multiple argocd apps inside the workflows-cluster"
```
garryod, line 3: Nit: Does this need quotes?

iamvigneshwars: No, it doesn't!

garryod (approved):
LGTM

PR #375, opened 2025-01-22
Add workflow status tests

yousefmoazzam:
Add tests for workflow queries where the workflow is in one of the following phases: "failed", "errored", "running", "null".

See the following github discussion regarding the "null" phase, as well as the enquiries we had to the argo workflow developers about getting a workflow in a "pending" phase: https://github.com/argoproj/argo-workflows/discussions/14119


garryod (commented):
LGTM. Only suggestion is that `test-resources` might be better named `test-assets`

garryod (approved):
LGTM

PR #376, opened 2025-01-23
Refactor workflow resolvers

yousefmoazzam:
A tentative attempt to do the refactoring mentioned in https://github.com/DiamondLightSource/workflows/pull/288#discussion_r1853857325

### DIFF
--- a/graph-proxy/src/graphql/workflows.rs
+++ b/graph-proxy/src/graphql/workflows.rs
@@ -11,6 +11,7 @@ use axum_extra::headers::{authorization::Bearer, Authorization};
 use chrono::{DateTime, Utc};
 use std::{collections::HashMap, ops::Deref, sync::Arc};
 use tracing::{debug, instrument};
+use url::Url;
 
 /// An error encountered when parsing the Argo Server API Workflow response
 #[derive(Debug, thiserror::Error)]
@@ -120,32 +121,11 @@ impl WorkflowRunningStatus {
 
     /// Tasks created by the workflow
     async fn tasks(&self, ctx: &Context<'_>) -> anyhow::Result<Vec<Task>> {
-        let mut url = ctx.data_unchecked::<ArgoServerUrl>().deref().to_owned();
-        let auth_token = ctx.data_unchecked::<Option<Authorization<Bearer>>>();
-        let mut nodes = self.0.status.as_ref().unwrap().nodes.clone();
-        if nodes.is_empty() {
-            url.path_segments_mut().unwrap().extend([
-                "api",
-                "v1",
-                "workflows",
-                self.0.as_ref().metadata.namespace.as_ref().unwrap(),
-                self.0.as_ref().metadata.name.as_ref().unwrap(),
-            ]);
-            let request = if let Some(auth_token) = auth_token {
-                CLIENT.get(url).bearer_auth(auth_token.token())
-            } else {
-                CLIENT.get(url)
-            };
-            nodes = request
-                .send()
-                .await?
-                .json::<APIResult<argo_workflows_openapi::IoArgoprojWorkflowV1alpha1Workflow>>()
-                .await?
-                .into_result()?
-                .status
-                .unwrap() //  Safe as the status field is always present
-                .nodes;
-        }
+        let url = ctx.data_unchecked::<ArgoServerUrl>().deref().to_owned();
+        let token = ctx
+            .data_unchecked::<Option<Authorization<Bearer>>>()
+            .to_owned();
+        let nodes = fetch_missing_task_info(url, token, Arc::clone(&self.0)).await?;
         Ok(TaskMap(nodes).into_tasks())
     }
 }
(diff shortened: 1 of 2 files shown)


graph-proxy/src/graphql/workflows.rs
```diff
@@ -30,35 +30,24 @@ pub(super) enum WorkflowParsingError {
...
+    async fn name(&self) -> &str {
+        self.0.metadata.name.as_ref().unwrap()
+    }
+
     /// The visit the Workflow was run against
-    visit: Visit,
-}
+    async fn visit(&self) -> &str {
+        self.0.metadata.namespace.as_ref().unwrap()
+    }
```
garryod, line 46: The response type here has changed from the `Visit` object to `str`

graph-proxy/src/graphql/workflows.rs
```diff
@@ -74,71 +63,65 @@ enum WorkflowStatus {
...
-        value: IoArgoprojWorkflowV1alpha1WorkflowStatus,
-        metadata: Arc<Metadata>,
+        workflow: Arc<IoArgoprojWorkflowV1alpha1Workflow>,
     ) -> Result<Option<Self>, WorkflowParsingError> {
-        match value.phase.as_deref() {
-            Some("Pending") => Ok(Some(Self::Pending(WorkflowPendingStatus::from(value)))),
-            Some("Running") => Ok(Some(Self::Running(WorkflowRunningStatus::new(
-                value, metadata,
-            )?))),
+        let status = workflow.status.as_ref().unwrap();
```
garryod, line 70: Instead of calling `unwrap` here could we return a variant of `WorkflowParsingError` as `unwrap` will cause the thread to panic?

Could we also pass the `status` to the `Workflow{Pending,Running,Complete}Status` constructors and avoid their `unwrap`s? You may need to `clone` the `status`, but it should be possible to pass the reference to it with an appropriate lifetime

garryod (changes requested):

graph-proxy/src/graphql/workflows.rs
yousefmoazzam: First one - done.

Second one - was a fun journey to do this; what I've done is likely not perfect, but it's in the ballpark of what was suggested I think.

yousefmoazzam:
I don't fully understand the lifetime elision rule(s) being applied to allow omission of the named lifetime annotation on the `impl` blocks, but I don't think it's worth me blocking review for that. I pinky promise to do some reading so then I understand it soon... :sweat_smile:

graph-proxy/src/graphql/workflows.rs
```diff
@@ -27,175 +28,221 @@ pub(super) enum WorkflowParsingError {
...
     }
 }
 
+#[derive(Debug)]
+pub(super) struct Metadata {
+    /// The name given to the workflow, unique within a given visit
+    name: String,
+    /// The visit the Workflow was run against
+    visit: Visit,
+}
```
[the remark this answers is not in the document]
yousefmoazzam: Yep, as noted in-person, we can try to make the `Metadata` get its info via references to the `IoArgoprojWorkflowV1alpha1Workflow` that `Workflow` has - but we'll do it separate to this PR, as a little side-quest :grin:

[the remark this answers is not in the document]
yousefmoazzam: Noted in #404

graph-proxy/src/graphql/workflows.rs
```diff
@@ -27,175 +28,221 @@ pub(super) enum WorkflowParsingError {
...
-#[derive(Debug, SimpleObject)]
+#[derive(Debug)]
 pub(super) struct Workflow {
-    /// Metadata containing name, proposal code, proposal number and visit of a workflow
-    #[graphql(flatten)]
-    metadata: Arc<Metadata>,
-    /// The time at which the workflow began running
-    status: Option<WorkflowStatus>,
+    /// Manifest associated with the workflow
+    pub(super) manifest: Arc<IoArgoprojWorkflowV1alpha1Workflow>,
```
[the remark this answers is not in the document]
yousefmoazzam: We can indeed, and I now understand why, thanks!

graph-proxy/src/graphql/workflows.rs
```diff
@@ -27,175 +28,221 @@ pub(super) enum WorkflowParsingError {
...
+    Running(WorkflowRunningStatus<'a>),
+    Succeeded(WorkflowSucceededStatus<'a>),
+    Failed(WorkflowFailedStatus<'a>),
+    Errored(WorkflowErroredStatus<'a>),
 }
 
-impl WorkflowStatus {
-    /// Creates a new `WorkflowStatus` from `IoArgoprojWorkflowV1alpha1WorkflowStatus` and associated metadata.
+impl<'a> WorkflowStatus<'a> {
+    /// Creates a new [`WorkflowStatus`] from [`IoArgoprojWorkflowV1alpha1Workflow`]
```
garryod, line 93: This still uses associated metadata, the comment can be left as is

graph-proxy/src/graphql/workflows.rs
```diff
@@ -27,175 +28,221 @@ pub(super) enum WorkflowParsingError {
...
-    fn from(value: IoArgoprojWorkflowV1alpha1WorkflowStatus) -> Self {
-        Self {
-            message: value.message,
-        }
+#[Object]
+impl WorkflowPendingStatus<'_> {
+    /// A human readable message indicating details about why the workflow is in this condition
+    async fn message(&self) -> Option<String> {
+        self.0.message.clone()
     }
```
garryod, line 143: Instead of cloning the message we could pass a ref
```suggestion
    async fn message(&self) -> Option<&str> {
        self.0.message.as_ref()
    }
```

garryod (changes requested):
Looking good, just a couple tiny things that could be tidied up

graph-proxy/src/graphql/workflows.rs
yousefmoazzam: Good catch, I do the same thing in a few other workflow status structs

PR #381, opened 2025-01-24
Use reusable workflows for supergraph CI

garryod:


.github/workflows/ci.yaml
```diff
@@ -49,24 +49,27 @@ jobs:
...
-    if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags')
-    needs: supergraph_composition
-    uses: ./.github/workflows/_supergraph_update.yaml
-    secrets:
-      GRAPH_FEDERATOR: ${{ secrets.GRAPH_FEDERATOR }}
-    
+    runs-on: ubuntu-latest
+    steps:
+      - name: Update Supergraph
+        uses: diamondlightsource/graph-federation@main
```
GDYendell, line 59: When do we recommend using `v1` vs `main`?

GDYendell (approved):

.github/workflows/ci.yaml
garryod: We should probably actually make them all `main` right now as we've not yet cut a release

garryod: Have changed them all to `main`for now

PR #391, opened 2025-01-27
Dev values update part 1

MaxHerbs:
Updated the dev-values for argocd, dashboard, events, and graph-proxy to work in a local workspace.

Are the workflow-cluster resources too limited? Particularly the controller. (See [charts/workflows-cluster/dev-values.yaml](https://github.com/DiamondLightSource/workflows/compare/dev-values?expand=1#diff-0025116216e9e07f8d592e4429e8bfbe06a12f67c369d867e7efdb885b5eb23c))

All other services are disabled

### DIFF
--- a/charts/apps/dev-values.yaml
+++ b/charts/apps/dev-values.yaml
@@ -1,5 +1,7 @@
 argocd:
   enabled: true
+  extraValueFiles:
+    - dev-values.yaml
 
 events:
   enabled: true
--- a/charts/apps/dev-values.yaml
+++ b/charts/apps/dev-values.yaml
@@ -1,3 +1,6 @@
+argocd:
+  enabled: true
+
 events:
   enabled: true
   extraValueFiles:
--- a/charts/apps/dev-values.yaml
+++ b/charts/apps/dev-values.yaml
@@ -52,6 +52,8 @@ workflows:
 
 graphProxy:
   enabled: true
+  extraValueFiles:
+    - dev-values.yaml
 
 dashboard:
   enabled: true
--- a/charts/apps/dev-values.yaml
+++ b/charts/apps/dev-values.yaml
@@ -62,3 +62,5 @@ graphProxy:
 
 dashboard:
   enabled: true
+  extraValueFiles:
+    - dev-values.yaml
--- a/charts/workflows-cluster/templates/ingress-tls-secret.yaml
+++ b/charts/workflows-cluster/templates/ingress-tls-secret.yaml
@@ -1,3 +1,4 @@
+{{- if .Values.ingress.enabled }}
 apiVersion: bitnami.com/v1alpha1
 kind: SealedSecret
 metadata:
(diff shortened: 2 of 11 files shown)


charts/apps/dev-values.yaml
```diff
@@ -46,12 +51,16 @@ sealedsecrets:
         memory: 256Mi
 
 workflows:
-  enabled: true
+  enabled: false
```
garryod, line 54: This can remain enabled

charts/apps/dev-values.yaml
```diff
@@ -30,12 +35,12 @@ kyverno:
...
 
 sessionspaces:
-  enabled: true
+  enabled: false
   extraValueFiles:
     - dev-values.yaml
 
 sealedsecrets:
-  enabled: true
+  enabled: false
```
garryod, line 43: This can remain enabled

charts/apps/dev-values.yaml
```diff
@@ -30,12 +35,12 @@ kyverno:
           memory: 1Gi
 
 sessionspaces:
-  enabled: true
+  enabled: false
```
garryod, line 38: This can remain enabled

charts/apps/dev-values.yaml
```diff
@@ -1,13 +1,18 @@
...
+  extraValueFiles:
+    - dev-values.yaml
+
 groups:
-  enabled: true
+  enabled: false
 
 kyverno:
-  enabled: true
+  enabled: false
```
garryod, line 15: This can remain enabled

charts/workflows-cluster/templates/ingress-tls-secret.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if .Values.ingress.enabled }}
```
garryod, line 1: Dunno how I managed to not add this condition despite having the option in the `values.yaml`

charts/apps/dev-values.yaml
```diff
@@ -1,13 +1,18 @@
...
     - dev-values.yaml
 
+events:
+  enabled: false
+  extraValueFiles:
+    - dev-values.yaml
+
 groups:
-  enabled: true
+  enabled: false
```
garryod, line 12: This can remain enabled

garryod (changes requested):
Looking good, only real change is that I'd rather the apps stayed enabled even if they're not working at the moment - a loud failure is better than silent

garryod (approved):
LGTM

garryod (changes requested):
Any chance you could fix some of the commit messages, then we can merge

MaxHerbs:
I've ammended some of the commits, and squashed a couple. Let me know if there's any further changes to make

garryod (approved):
LGTM

iamvigneshwars (approved):

PR #394, opened 2025-01-28
Update argo workflows to v3.6.2

### LINKED ISSUE #393: Can't submit workflows with Argo Workflows UI
The Argo workflows UI ([https://argo-workflows.workflows.diamond.ac.uk](https://argo-workflows.workflows.diamond.ac.uk)) currently does not allow to submit a workflow from the Cluster Workflow Templates tab. The namespace seems to be set as "undefined". 

Resubmitting existing workflows works fine (because namespace is already stored in the workflow) and submitting workflows with `kubectl` also works fine.

daurer:
This was an attempt to fix #393 which unfortunately did not resolve the issue, but might as well update to the latest version of Argo workflows.

### DIFF
--- a/charts/workflows/Chart.yaml
+++ b/charts/workflows/Chart.yaml
@@ -3,12 +3,12 @@ name: workflows
 description: Data Analysis workflow orchestration
 type: application
 
-version: 0.13.0
+version: 0.13.1
 
 dependencies:
   - name: argo-workflows
     repository: https://argoproj.github.io/argo-helm
-    version: 0.45.1
+    version: 0.45.4
     condition: argo-workflows.enabled
   - name: postgresql-ha
     repository: https://charts.bitnami.com/bitnami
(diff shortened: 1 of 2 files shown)


garryod (approved):
LGTM

charts/workflows/Chart.yaml
```diff
@@ -3,12 +3,13 @@ name: workflows
 description: Data Analysis workflow orchestration
 type: application
 
-version: 0.13.0
+
```
garryod, line 6: Why the bonus whitespace?

garryod (changes requested):
Looks good, bar the additional whitespace

charts/workflows/Chart.yaml
daurer: Lack of attention to detail :smiley:

garryod (approved):

PR #395, opened 2025-01-29
Update devcontainer rust version to 1.84.0

yousefmoazzam:
To match the rust version the CI uses


garryod:
Could we set up `dependabot` to do this for us? Should just be an addition of directories [here](https://github.com/DiamondLightSource/workflows/blob/a49343f1af2b446387d838c1e7220752ed6e92e0/.github/dependabot.yaml#L17)

yousefmoazzam:
Nice, that sounds better to automate it, am happy to take a look at doing that.

PR #397, opened 2025-01-29
add OIDC to argocd

iamvigneshwars:


charts/argocd/templates/argocd-oauth-secret.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: bitnami.com/v1alpha1
+kind: SealedSecret
+metadata:
+  labels:
+    app.kubernetes.io/part-of: argocd
```
garryod, line 5: Probably don't need this on the `sealedsecret`

charts/argocd/templates/argocd-oauth-secret.yaml
```diff
@@ -0,0 +1,20 @@
...
+  annotations:
+    sealedsecrets.bitnami.com/namespace-wide: "true"
+  name: argocd-oauth
+  namespace: argocd
```
garryod, line 9: This should be templated

charts/argocd/templates/argocd-oauth-secret.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: bitnami.com/v1alpha1
+kind: SealedSecret
+metadata:
+  labels:
+    app.kubernetes.io/part-of: argocd
+  annotations:
+    sealedsecrets.bitnami.com/namespace-wide: "true"
+  name: argocd-oauth
```
garryod, line 8: This should be templated with the `common.names.fullname` as a prefix

charts/argocd/values.yaml
```diff
@@ -1,5 +1,32 @@
...
+    cm:
+      url: https://argo-cd.workflows.diamond.ac.uk
+      oidc.config: |
+        name: Keycloak
+        issuer: https://authn.diamond.ac.uk/realms/master
+        clientID: workflows-argocd
+        clientSecret: $argocd-oauth:oidc.keycloak.clientSecret
+        requestedIDTokenClaims:
+          groups:
+            essential: true
```
garryod, line 13: Do we need groups? We're not going to be using them for our AuthZ afaik

charts/argocd/templates/argocd-oauth-secret.yaml
iamvigneshwars: The argocd configmap cannot template it

garryod: Gotcha, did you stick in an upsteam PR to add `tpl`?

PR #405, opened 2025-01-30
add team members to argocd admins

iamvigneshwars:


charts/argocd/values.yaml
```diff
@@ -24,6 +26,9 @@ argo-cd:
         p, role:admin, logs, get, */*, allow
         p, role:admin, settings, *, *, allow
         g, admin, role:admin
+        g, benedikt.daurer@diamond.ac.uk, role:admin
+        g, garry.o'donnell@diamond.ac.uk, role:admin
+        g, vigneshwar.shanmugasundar@diamond.ac.uk, role:admin
```
garryod, line 31: Can you add Max & Thomas too?

iamvigneshwars: Ahh, forgot to add them!

garryod (approved):

PR #406, opened 2025-01-30
add telemetry to sessionspaces

iamvigneshwars:


sessionspaces/src/telemetry.rs
garryod, line 1: Would it be worth pulling this out into a library? Since it seems to be identical to the `graph-proxy` `telemetry.rs`?

PR #409, opened 2025-01-31
add telemetry library

iamvigneshwars:


sessionspaces/Dockerfile
```diff
@@ -6,7 +6,8 @@ WORKDIR /app
 
 RUN cargo install cargo-auditable
 
-COPY Cargo.toml Cargo.lock ./
+COPY sessionspaces/Cargo.toml sessionspaces/Cargo.lock ./
+COPY telemetry /telemetry
```
garryod, line 10: Do we want to do a `pre-build` step with the `telemetry` crate too?

graph-proxy/Dockerfile
garryod, line 1: It's a bit odd having the `Dockerfile` in the wrong context for it's execution, is there a better way we can go about this?

garryod (changes requested):

backend/graph-proxy/Cargo.toml
```diff
@@ -18,24 +18,19 @@ derive_more = { workspace = true }
...
-tracing-opentelemetry = { version = "0.28.0" }
-tracing-subscriber = { workspace = true }
 url = { workspace = true }
 
 [dev-dependencies]
 mockito = "1.6.1"
 
 [build-dependencies]
-built = { version = "0.7.5" }
+built = { workspace = true }
```
garryod, line 36: Do we need built here still? I think it was only being used for telemetry stuff

backend/telemetry/src/lib.rs
```diff
@@ -31,23 +46,36 @@ impl Drop for OtelGuard {
...
+    /// The endpoint to send OTLP traces to
+    #[arg(short, long, env = "TRACING_ENDPOINT")]
+    pub tracing_endpoint: Option<Url>,
+    /// The minimum telemetry level
+    #[arg(short, long, env="TELEMETRY_LEVEL", default_value_t=Level::INFO)]
+    pub telemetry_level: Level,
+}
+
 /// Sets up Logging & Tracing using opentelemetry if available
 pub fn setup_telemetry(
```
garryod, line 74: Would it be worth having this method take `TelemetryArgs`? Also, might be worth calling it `TelemetryConfig` and just having it implement `clap::Parser`

backend/Dockerfile.sessionspaces
```diff
@@ -9,21 +9,27 @@ RUN cargo install cargo-auditable
 COPY argo-workflows-openapi/Cargo.toml argo-workflows-openapi/Cargo.toml
 COPY graph-proxy/Cargo.toml graph-proxy/
 COPY sessionspaces/Cargo.toml sessionspaces/
+COPY telemetry/Cargo.toml telemetry/Cargo.toml
+COPY telemetry/build.rs telemetry/build.rs
+COPY telemetry/src/lib.rs telemetry/src/lib.rs
```
garryod, line 14: Can we avoid copying in the telemetry source here and instead leave that until the final build?

backend/Dockerfile.graph-proxy
```diff
@@ -12,20 +12,25 @@ COPY argo-workflows-openapi/build.rs argo-workflows-openapi/build.rs
 COPY argo-workflows-openapi/src/lib.rs argo-workflows-openapi/src/lib.rs
 COPY sessionspaces/Cargo.toml sessionspaces/Cargo.toml
 COPY graph-proxy/Cargo.toml graph-proxy/
+COPY telemetry/Cargo.toml telemetry/Cargo.toml
+COPY telemetry/build.rs telemetry/build.rs
+COPY telemetry/src/lib.rs telemetry/src/lib.rs
```
garryod, line 17: Can we avoid copying in the telemetry source here?

garryod (changes requested):
Looking good, just a few little bits and pieces

backend/telemetry/src/lib.rs
iamvigneshwars: Yeah, we can do that

backend/sessionspaces/src/main.rs
```diff
@@ -35,19 +36,17 @@ struct Cli {
...
 
 #[tokio::main(flavor = "current_thread")]
 async fn main() {
     dotenvy::dotenv().ok();
     let args = Cli::parse();
 
-    tracing_subscriber::fmt()
-        .with_max_level(args.log_level)
-        .init();
+    let _otlp_guard = setup_telemetry(args.telemetry_config.clone()).unwrap();
```
garryod, line 49: Do we need to `clone` this?

backend/telemetry/src/lib.rs
```diff
@@ -19,6 +20,20 @@ mod built_info {
...
+    MetricExporterBuildError(String),
+    #[error("Span Exporter Build error: {0}")]
+    SpanExporterBuildError(String),
+    #[error("Tracing configuration error: {0}")]
+    TracingConfigError(String),
+    #[error("Tracer shutdown error: {0}")]
+    TracerShutdownError(String),
+    #[error("Meter shutdown error: {0}")]
+    MeterShutdownError(String),
+}
```
garryod, line 35: Why are we converting these to strings? Could we instead just capture the underlying error types? E.g.

```rust
#[error("Could not create MetricExporter: {0}")]
MetricExporterBuildError(#[from] opentelemetry_sdk::metrics::MetricError)
```

The `#[from]` here will also allow us to use `Try` (`?`)

backend/Dockerfile.graph-proxy
```diff
@@ -12,20 +12,26 @@ COPY argo-workflows-openapi/build.rs argo-workflows-openapi/build.rs
 COPY argo-workflows-openapi/src/lib.rs argo-workflows-openapi/src/lib.rs
 COPY sessionspaces/Cargo.toml sessionspaces/Cargo.toml
 COPY graph-proxy/Cargo.toml graph-proxy/
+COPY telemetry/build.rs telemetry/build.rs
+COPY telemetry/Cargo.toml telemetry/Cargo.toml
```
garryod, line 16: These could be a single COPY

garryod (changes requested):

backend/sessionspaces/src/main.rs
iamvigneshwars: Ah, lost in copy and paste from `graph-proxy`

backend/telemetry/src/lib.rs
iamvigneshwars: Yeah, could do that!

backend/telemetry/src/lib.rs
```diff
@@ -19,6 +20,20 @@ mod built_info {
...
+    MetricExporterBuildError(opentelemetry_sdk::metrics::MetricError),
+    #[error("Span Exporter Build error: {0}")]
+    SpanExporterBuildError(#[from] opentelemetry::trace::TraceError),
+    #[error("Tracing Subscriber initialization error: {0}")]
+    TracingSubscriberInitError(#[from] tracing_subscriber::util::TryInitError),
+    #[error("Tracer shutdown error: {0}")]
+    TracerShutdownError(opentelemetry::trace::TraceError),
+    #[error("Meter shutdown error: {0}")]
+    MeterShutdownError(#[from] opentelemetry_sdk::metrics::MetricError),
+}
```
garryod, line 35: It's probably worth de-duplicating the common error types, they should contain enough information within to tell if it's from build or shutdown

garryod (changes requested):

garryod (approved):
LGTM

PR #411, opened 2025-01-31
Make Rust binaries scannable with `cargo-auditable`

garryod:
Compiles rust binaries in containers using `cargo-auditable` as described [in the dev portal](https://dev-portal.diamond.ac.uk/guide/rust/how-tos/auditable-builds/)


iamvigneshwars (approved):
LGTM

PR #417, opened 2025-02-03
Update Rust service containers to use stable debian build images

garryod:
Not sure why we were using `bullseye` before, have been using `bookworm` for our dev-containers forever


MaxHerbs (approved):
LGTM

PR #421, opened 2025-02-04
Add an ArgoCD cargo to the dashboard

garryod:


frontend/dashboard/src/components/dashboard.tsx
```diff
@@ -122,6 +140,9 @@ function Dashboard() {
           <Grid key={"grafana"} size={{ xs: 2, sm: 2, md: 3 }}>
             <Card style={{ height: "100%" }}>{grafanaCard}</Card>
           </Grid>
+          <Grid key={"grafana"} size={{ xs: 2, sm: 2, md: 3 }}>
+            <Card style={{ height: "100%" }}>{argoCdCard}</Card>
+          </Grid>
```
daurer, line 145:
```suggestion
          <Grid key={"argocd"} size={{ xs: 2, sm: 2, md: 3 }}>
            <Card style={{ height: "100%" }}>{argoCdCard}</Card>
          </Grid>
```

daurer (changes requested):

frontend/dashboard/src/components/dashboard.tsx
garryod: Good catch

PR #426, opened 2025-02-11
Dev values CI

MaxHerbs:


iamvigneshwars:
I was tempted to used the copilot 😅 
Overall it looks good, apart from those typos

charts/workflows/dev-values.yaml
```diff
@@ -36,12 +36,12 @@ argo-workflows:
...
-        cpu: 100m
-        memory: 256Mi
+        cpu: 500m
+        memory: 2Gi
         ephemeral-storage: 256Mi
       requests:
-        cpu: 50m
-        memory: 128Mi
+        cpu: 500m
+        memory: 2500m
```
iamvigneshwars, line 44: Its still `2500m` for memory

iamvigneshwars (changes requested):

charts/workflows/dev-values.yaml
MaxHerbs: Yes, the dev namespaces have 40Gb so this is still well within quota

PR #433, opened 2025-02-17
add test to telemetry library

iamvigneshwars:
This add unit test to telemetry library to test if the metrics and traces are properly exported. 
I've used `tonic` to mock the gRPC services as the `otel-collector` uses the same.

### DIFF
--- a/.github/workflows/_telemetry_code.yaml
+++ b/.github/workflows/_telemetry_code.yaml
@@ -1,0 +1,70 @@
+name: Telemetry Code
+
+on:
+  workflow_call:
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout source
+        uses: actions/checkout@v4.2.2
+
+      - name: Install stable toolchain
+        uses: actions-rust-lang/setup-rust-toolchain@v1.10.1
+        with:
+          cache: false
+          components: clippy,rustfmt
+
+      - name: Cache Rust Build
+        uses: Swatinem/rust-cache@v2.7.7
+        with:
+          workspaces: backend/telemetry
+          shared-key: backend/telemetry
+
+      - name: Check Formatting
+        working-directory: backend/telemetry
+        run: >
+          cargo fmt
+          --check
+
+      - name: Lint with Clippy
+        working-directory: backend/telemetry
+        run: >
+          cargo clippy
+          --all-targets
+          --all-features
+          --no-deps
+          --
+          --deny warnings
+
+      - name: Check Dependencies with Cargo Deny
+        uses: EmbarkStudios/cargo-deny-action@v2.0.4
+        with:
+          command: check licenses ban
+          manifest-path: backend/Cargo.toml
+
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout source
+        uses: actions/checkout@v4.2.2
+
+      - name: Install stable toolchain
+        uses: actions-rust-lang/setup-rust-toolchain@v1.10.1
+        with:
+          cache: false
+          components: rustfmt
+
+      - name: Cache Rust Build
+        uses: Swatinem/rust-cache@v2.7.7
+        with:
+          workspaces: backend/telemetry
+          shared-key: backend/telemetry
+
+      - name: Run Tests
+        working-directory: backend/telemetry
+        run: >
+          cargo test
+          --all-targets
+          --all-features
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -104,3 +104,8 @@ jobs:
     permissions:
       contents: write
       pull-requests: write
+  
+  telemetry_code:
+    # Deduplicate jobs from pull requests and branch pushes within the same repo.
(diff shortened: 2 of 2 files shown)


backend/telemetry/Cargo.toml
```diff
@@ -18,6 +18,10 @@ tracing-opentelemetry = { version = "0.28.0" }
 tracing-subscriber = { version = "0.3.19" }
 url = { workspace = true }
 
+[dev-dependencies]
+tokio = { workspace = true }
+opentelemetry-proto = { version = "0.28.0" }
+tonic = { version = "0.12.3" }
```
yousefmoazzam, line 24: Another nitpick, take it or leave it: seems like it's general convention in `Cargo.toml` files to order the crates alphabetically

backend/telemetry/src/lib.rs
```diff
@@ -122,3 +122,89 @@ pub fn setup_telemetry(config: TelemetryConfig) -> Result<OtelGuard, TelemetryEr
...
+        async fn export(
+            &self,
+            request: Request<ExportTraceServiceRequest>,
+        ) -> Result<Response<ExportTraceServiceResponse>, Status> {
+            self.requests.lock().unwrap().push(request.into_inner());
+            Ok(Response::new(ExportTraceServiceResponse::default()))
+        }
+    }
+
+    #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
```
yousefmoazzam, line 174: This is a curiosity more than a code review point: I can see that using the default current-thread runtime via `#[tokio::test]` causes the `Drop` implementation for `OtelGuard` to hang when trying to shutdown the "tracer provider", but I don't understand the code enough to know why this is the case.

My very limited understanding is that the test is spawning a task with this "telemetry server" running inside it,  doing minimal stuff to try and get a metric and a trace emitted, and then asserting that a metric and trace were indeed emitted.

Dumb question: why does the task with the "telemetry server" running inside it need to be on a separate thread to the test's thread, in order for the tracer provider to not get stuck shutting down?

Apologies if the question doesn't make sense, feel free to disregard, like I said, am just curious to understand how this works!

yousefmoazzam (commented):
Looks good, just a few nitpicks, and a question purely out of curiosity if you're willing to humour me!

backend/telemetry/src/lib.rs
```diff
@@ -122,3 +122,89 @@ pub fn setup_telemetry(config: TelemetryConfig) -> Result<OtelGuard, TelemetryEr
         .try_init()?;
     Ok(otel_guard)
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
```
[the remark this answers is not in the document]
iamvigneshwars: Yeah, good catch

backend/telemetry/src/lib.rs
```diff
@@ -122,3 +122,89 @@ pub fn setup_telemetry(config: TelemetryConfig) -> Result<OtelGuard, TelemetryEr
...
+            metrics_endpoint: Some(Url::parse("http://[::1]:50051").unwrap()),
+            tracing_endpoint: Some(Url::parse("http://[::1]:50051").unwrap()),
+            telemetry_level: Level::INFO,
+        };
+        {
+            let _guard = setup_telemetry(config).unwrap();
+            tracing::info!(monotonic_counter.test_metric = 1u64, "Test metric");
+            let span = tracing::info_span!("test_span");
+            let _enter = span.enter();
+        }
```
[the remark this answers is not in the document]
iamvigneshwars: We usually enter the span to activate it, but yeah just creating the span is sufficient for this test

backend/telemetry/src/lib.rs
iamvigneshwars: Yeah you are right! So the `shutdown` on `meter_provider` and `tracer_provider` tries to flush any remaining mertics and spans, and waits for some internal worker tasks to finish before returning.

With the default runtime via `#[tokio::test]` there will be deadlock, as it's gonna wait for some tasks thats never gonna be executed with the current-thread runtime. 

We run the "telemetry server" on a seperate thread simple because it blocks the current thread. When we call `serve()` it would run indefenitely and block the current execution.

backend/telemetry/src/lib.rs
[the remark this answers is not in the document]
yousefmoazzam: Ok cool, as long as it doesn't affect what was being tested for, that's alright.

backend/telemetry/src/lib.rs
yousefmoazzam: > Yeah you are right! So the `shutdown` on `meter_provider` and `tracer_provider` tries to flush any remaining mertics and spans, and waits for some internal worker tasks/threads to finish before returning.
> 
> With the default runtime via `#[tokio::test]` there will be deadlock, as it's gonna wait for some tasks thats never gonna be executed with the current-thread runtime.


Ah ok, that's starting to make some sense to me, the `shutdown` requires some tasks to finish, but they can't if using the current-thread runtime.

I'm still not 100% there, but that's fine, that's likely just me being slow, thanks for the explanation! :slightly_smiling_face:

yousefmoazzam (approved):

PR #442, opened 2025-02-24
fix(dev-values): modified dev values for apps, sessionspaces and argocd

TBThomas56:
Removed sealed secrets and ingresses

### DIFF
--- a/charts/sessionspaces/Chart.yaml
+++ b/charts/sessionspaces/Chart.yaml
@@ -3,7 +3,7 @@ name: sessionspaces
 description: Namespace controller for creating session namespaces
 type: application
 
-version: 0.3.5
+version: 0.3.6
 appVersion: 0.1.0-rc25
 dependencies:
   - name: common
--- a/charts/sessionspaces/Chart.yaml
+++ b/charts/sessionspaces/Chart.yaml
@@ -9,3 +9,6 @@ dependencies:
   - name: common
     version: 2.23.0
     repository: oci://docker.io/bitnamicharts
+  - name: database
+    version: 0.1.0
+    repository: file://./charts/database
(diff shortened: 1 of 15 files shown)


charts/sessionspaces/charts/database/Chart.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v2
+name: database
+description: Ispyb database deployment
+type: application
+
+version: 0.1.0
+appVersion: 0.1.0-rc25
```
iamvigneshwars, line 7: why do we have `appVersion` here?

charts/sessionspaces/charts/database/templates/deployment.yaml
```diff
@@ -0,0 +1,49 @@
...
+            {{- $.Values.resources | toYaml | nindent 12 }}
+          env:
+            - name: MARIADB_ROOT_PASSWORD
+              value: {{ $.Values.password.value }}
+          {{- with $.Values.nodeSelector }}
+          nodeSelector:
+            {{- . | toYaml | nindent 8 }}
+          {{- end }}
+          ports:
+            - containerPort: {{ $.Values.dbPort }}
```
iamvigneshwars, line 40: duplicate `ports`  can be removed, also it's  common to have the `name`, `protocol`  and `targetPort` for the `ports`

charts/sessionspaces/Chart.yaml
```diff
@@ -3,9 +3,12 @@ name: sessionspaces
 description: Namespace controller for creating session namespaces
 type: application
 
-version: 0.3.5
+version: 0.3.7
```
iamvigneshwars, line 6: think the version should be `0.3.6` ?

charts/sessionspaces/charts/database/templates/deployment.yaml
MaxHerbs: Fixed, see new push

iamvigneshwars (approved):
LGTM

PR #449, opened 2025-02-27
Deploy base apps

MaxHerbs:
This is currently incomplete. See changes to name templating


MaxHerbs:
Templating `"{{ include "common.names.fullname" $ }}-authn-config"` in the values file isn't possible. I suggest we dont template it at all (at least in the short term) so that we are able to deploy the cluster with any name (we are currently stuck with only `workflows-cluster`). This should let us deploy with `base-apps` though argocd

PR #456, opened 2025-03-11
fix(frontend): prevent duplicate nodes & edges in react flow

VictoriaBeilsten-Edmands:
Check for matching ID for nodes and edges before adding to array. Calling add on a set does not work here as Node and Edge are not primitive types and are not compared by value.


iamvigneshwars (approved):
LGTM

PR #469, opened 2025-03-18
Workflows docs

MaxHerbs:
Docs for mounting tmp volume to workflow 
Start of docs for workflows CLI

### DIFF
--- a/docs/how-tos/submit-workflow.md
+++ b/docs/how-tos/submit-workflow.md
@@ -96,7 +96,7 @@ and login with your Diamond credentials. It should open the dashboard in the wor
     A workflow can only by submitted within a namespace. For now, the available namespaces are all of Diamond's visits. 
     In the future, we anticipate to also add other namespaces, e.g. user/fedid
 
-After specifying the namespace in which you would like your workflow to be executed, you can click on 
+After specifying the namespace in which you would like your workflow to be executed, you can click on
 ```+ SUBMIT NEW WORKFLOW``` which should give you a prompt
 that allows you to either select an existing template or edit a default template:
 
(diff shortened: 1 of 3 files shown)


docs/how-tos/submit-workflow.md
```diff
@@ -118,14 +118,34 @@ has successfully finished, it should look like this:
...
+awf list                       # list current workflows
+awf get hello-world-xxx        # get info about a specific workflow
+awf logs hello-world-xxx       # print the logs from a workflow
+awf delete hello-world-xxx     # delete workflow
+```
 
-    Since we currently rely on CAS for authentication, we cannot use the Argo CLI to submit a workflow.
-    
\ No newline at end of file
+When submitting workflows, please use your visit ID as the namespace, eg `mg36964-1`
```
daurer, line 151:
```suggestion
When submitting workflows, please use your instrumentSession ID as the namespace, eg `mg36964-1`
```

daurer (changes requested):
Looks good, just a few small suggestions. And maybe you could add this template example to the examples group?

examples/temporary-volume.yaml
```diff
@@ -0,0 +1,48 @@
+apiVersion: argoproj.io/v1alpha1
+kind: ClusterWorkflowTemplate
+metadata:
+  name: exec-bash
```
daurer, line 4:
```suggestion
  name: mount-tmpdir
```

daurer (changes requested):
maybe rename the template to something more descriptive than `exec-bash`

daurer (approved):
LGTM

PR #480, opened 2025-03-21
Populate mock sessionspaces

MaxHerbs:


iamvigneshwars:
We can use the `ispyb_build` database which is already populated

iamvigneshwars:
I don't think was need to explictly run the ` build.sh` just pointing the endpoint to `ispyb_build` should be fine 🤔 what's what we do in [devcontainer](https://github.com/DiamondLightSource/workflows/blob/main/backend/sessionspaces/.devcontainer/docker-compose.yml)

Did you try to simply point to `ispyb_build` ? If not maybe try it? 

I also be confusing it with something else....

iamvigneshwars (approved):

PR #481, opened 2025-03-21
Mock s3 update

MaxHerbs:
The current deployment has some issues with the service configuration for localstacks, and the charts are quite rigid making it difficult to correct. I have moved the dev deployment to adobe's s3mock container, and updated the charts accordingly.

### DIFF
--- a/charts/workflows/charts/s3mock/templates/deployment.yaml
+++ b/charts/workflows/charts/s3mock/templates/deployment.yaml
@@ -1,0 +1,56 @@
+{{- if $.Values.enabled }}
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ include "common.names.fullname" $ }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "common.labels.standard" $ | nindent 4 }}
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      {{- include "common.labels.matchLabels" $ | nindent 6 }}
+  template:
+    metadata:
+      {{- with $.Values.podAnnotations }}
+      annotations:
+        {{- . | toYaml | nindent 8 }}
+      {{- end }}
+      labels:
+        {{- include "common.labels.matchLabels" $ | nindent 8 }}
+    spec:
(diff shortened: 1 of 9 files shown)


charts/workflows/dev-values.yaml
```diff
@@ -1,18 +1,21 @@
...
-    awslocal s3api create-bucket --bucket k8s-workflows-test --region unsupported
-  replicaCount: 1
-  resource:
-    limits:
-      cpu: 100m
-      memory: 128Mi
+  image:
+    repository: adobe/s3mock
+    pullPolicy: Always
+    tag: "latest"
```
iamvigneshwars, line 6: We should pin a specific version here and should avoid using the `latest` tag

charts/workflows/charts/s3mock/templates/deployment.yaml
```diff
@@ -0,0 +1,56 @@
...
+      labels:
+        {{- include "common.labels.matchLabels" $ | nindent 8 }}
+    spec:
+      imagePullSecrets: {{ include "common.images.renderPullSecrets" (dict "images" (list $.Values.image) "context" $ ) }}
+      containers:
+        - name: s3mock
+          image: {{ include "common.images.image" ( dict "imageRoot" $.Values.image "global" $.Values "chart" $.Chart )}}
+          imagePullPolicy: {{ $.Values.image.pullPolicy }}
+          ports:
+            - name: bucket-port
```
iamvigneshwars, line 29: Just a nit, we usually won't add the parent name here, it would be just `bucket` or `mock-bucket`

charts/workflows/charts/s3mock/templates/deployment.yaml
```diff
@@ -0,0 +1,56 @@
...
+              protocol: TCP
+              targetPort: {{ $.Values.bucketPort }}
+              containerPort: {{ $.Values.bucketPort }}
+            - name: bucket-port-s
```
iamvigneshwars, line 33: Same as above

iamvigneshwars (changes requested):

MaxHerbs:
All above changes have been made. Let me know if there's anything else

iamvigneshwars (approved):
LGTM

PR #501, opened 2025-03-27
Workflows with dev-values

MaxHerbs:
This is quite a significant overhaul to the workflows engine in the dev deployment.
I have disabled auth etc for the vcluster and workflows, rather than mocking as the OIDC flow would require an ingress etc to be able to authenticate which isnt doable in the dev values.

Workflows is now running without any auth, and instead the UID should be set in values or over-written in the argo dashboard
Several tweaks have been made to the s3 config so it works with the mock credentials.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,8 @@ vcluster connect workflows-cluster --silent -- <COMMAND>
 helm install workflows-cluster charts/workflows-cluster -f charts/workflows-cluster/dev-values.yaml
 ```
 
+If you wish to run workflows, you should override the `uid` in the workflows app with your own uid.
+
 ## Serve Docs
 
 Firstly, install `mkdocs` and the requisite dependencies in `docs/requirements.txt`; For this you may wish to use `pipx`, as:
--- a/charts/workflows/dev-values.yaml
+++ b/charts/workflows/dev-values.yaml
@@ -12,10 +12,10 @@ s3mock:
   resources:
     requests:
       cpu: 100m
-      memory: 128Mi
+      memory: 1Gi
     limits:
       cpu: 200m
-      memory: 512Mi
+      memory: 1Gi
 
 argo-workflows:
   artifactRepository:
--- a/charts/workflows/dev-values.yaml
+++ b/charts/workflows/dev-values.yaml
@@ -44,6 +44,10 @@ argo-workflows:
         ephemeral-storage: 128Mi
   server:
     replicas: 1
+    authModes: ["server"]
+    extraEnv:
+      - name: ARGO_SECURE
+        value: "false"
     resources:
       limits:
         cpu: 100m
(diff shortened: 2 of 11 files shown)


charts/workflows-cluster/templates/authn-config-configmap.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if .Values.authenticationConfiguration.enabled }}
```
iamvigneshwars, line 1: Is this enabled in `values.yaml`?

charts/workflows/dev-values.yaml
```diff
@@ -47,6 +45,10 @@ argo-workflows:
         ephemeral-storage: 128Mi
   server:
     replicas: 1
+    authModes: ["server"]
+    extraEnv:
+      - name: ARGO_SECURE
+        value: "false"
```
iamvigneshwars, line 51: Just curious, does this turn off auth?

iamvigneshwars:
Just a nit, it's good to add a small description if the PR has many changes. It will be easy to review without any context.

charts/workflows/dev-values.yaml
MaxHerbs: Yes, this allows you to log into the argo-workflows dashboard without OIDC credentials or any kind of key from a service account etc

iamvigneshwars:
> Just a nit, Sometimes it's good to add a small description if the PR has many changes. It will be easy to review without any context.

Ahh, Nvm didn't realise you added description later 😅

MaxHerbs:
Don't worry, I had forgotten to until you prompted :smile:

charts/workflows-cluster/templates/authn-config-configmap.yaml
iamvigneshwars: That config map has the authn config for the vcluster, so we do need that?? 
I was just wondering since you added a condition there, if you forget to enable it in `values.yaml`

MaxHerbs: I mean we may not need to introduce the templating at all. In the dev values it should create an empty config map which is then never used, so adding an option to disable it may be a little redundant

MaxHerbs: I'm going to drop that commit. It should create the config map with the same values as prod, but never mount the config map to the cluster. Not worth introducing more complexity by adding complexity

iamvigneshwars (dismissed):
LGTM

README.md
```diff
@@ -27,6 +27,8 @@ vcluster connect workflows-cluster --silent -- <COMMAND>
 helm install workflows-cluster charts/workflows-cluster -f charts/workflows-cluster/dev-values.yaml
 ```
 
+If you wish to run workflows , you should override the `uid` in the workflows app with your own uid.
```
iamvigneshwars, line 30: There should be no space before `,`

iamvigneshwars (changes requested):

iamvigneshwars (approved):

PR #502, opened 2025-03-27
Workflow template ref on graph

MaxHerbs:
Added the workflow_template_ref endpoint to the graph
Added a testcase for the endpoint
Removed some broken CLI config that only works in release mode


iamvigneshwars (approved):
LGTM

PR #518, opened 2025-04-07
Filter workflows based on their status

iamvigneshwars:


backend/graph-proxy/src/graphql/workflows.rs
```diff
@@ -527,12 +570,21 @@ impl WorkflowsQuery {
         visit: VisitInput,
         cursor: Option<String>,
         #[graphql(validator(minimum = 1, maximum = 10))] limit: Option<u32>,
+        #[graphql(default)] filter: WorkflowStatusFilter,
```
MaxHerbs, line 573: Is 'filter' the best name for this?

iamvigneshwars: Yes, its the widely accepted name in graphql, we can expand this if we add more filters in the future

MaxHerbs: How do we then extend this `filter` for other filters that we need? eg the creator of the workflow

backend/graph-proxy/src/graphql/workflows.rs
```diff
@@ -1342,4 +1394,83 @@ mod tests {
...
+                nodes {{
+                    name
+                }}
+            }}
+        }}
+    "#,
+            visit.proposal_code, visit.proposal_number, visit.number,
+        );
+        schema.execute(query).await.into_result().unwrap();
+        workflows_endpoint.assert_async().await;
```
yousefmoazzam, line 1444: I noticed that there's an assertion that the mock endpoint gets hit with the expected request and query params, but no assertion on the data returned by the graphql service.

Would it be worth checking the data returned by the graphql service is indeed filtered, just in case there's some issue with the resolvers somehow incorrectly dealing with the filtered responses from the argo workflows server? Or do you think would that be overkill?

backend/graph-proxy/src/graphql/workflows.rs
```diff
@@ -1342,4 +1394,83 @@ mod tests {
...
+            workflows(
+                visit: {{proposalCode: "{}", proposalNumber: {}, number: {}}}, 
+                filter: {{ error: true }}
+            ) {{
+                nodes {{
+                    name
+                }}
+            }}
+        }}
+    "#,
```
yousefmoazzam, line 1440: Nitpick: the indentation level of the code starting with `query` in the raw string literal looks a bit odd to me.

I know `rustfmt` doesn't fiddle with code involving raw string literals much so I could see how this doesn't get picked up by the code formatter, but personally, I'd prefer it indented so that code within a function doesn't go to the left as if it's not in the scope of the function (which the last line of the raw string literal does), if you see what I mean.

yousefmoazzam (commented):
Overall looks good! I agree with @MaxHerbs in the other thread that it'd be good to clarify how it'd work to extend to more filters.

One question, and a nitpick which you're free to take or leave.

backend/graph-proxy/src/graphql/workflows.rs
iamvigneshwars: Discussed in-person

backend/graph-proxy/src/graphql/workflows.rs
iamvigneshwars: Discussed in-person

MaxHerbs (approved):

PR #519, opened 2025-04-08
Filter creator

MaxHerbs:
This is a bit rough still but its just a draft, I'm looking for some input on the filter implementation.
This is a branch off Viki's branch - see only the most recent commit

Alternatively I could move the filtering logic into the specific filter struct, and then iterate through all `Some` filters in the parent WorkflowFilter, and minimise the code in the WorkflowFilter.generate_filter


iamvigneshwars (approved):
LGTM

PR #520, opened 2025-04-09
Graph parameters

MaxHerbs:
Added workflow parameters to the graph
Added test for parameters

The workflow seems to provide an empty list, rather than none where there are no parameters. Is an empty list desirable, or should the graph check len>0 and return none if not?


iamvigneshwars (approved):
LGTM

PR #523, opened 2025-04-10
replace gprc with http for telemetry

iamvigneshwars:
Since we are gonna use ingress to send telemetry outside the vitrual cluster, our ingress controller is not handling grpc to http (HTTP/2 vs HTTP/1)  conversion properly. It's worth replacing our telemtry with http since we are using ingress anyway.


MaxHerbs (approved):
LGTM

PR #524, opened 2025-04-10
docs: creating artefacts

MaxHerbs:


docs/how-tos/create-artefacts.md
```diff
@@ -0,0 +1,195 @@
...
+      parameters:
+      - name: command
+    container:
+      image: busybox
+      volumeMounts:
+      - name: tmpdir
+        mountPath: /tmp
+      command: ["/bin/sh", "-c"]
+      args: - "{{ inputs.parameters.command }}"
+  #vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv
```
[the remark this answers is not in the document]
yousefmoazzam: Oh I see I think, it's talking about this chunk of YAML:
```yaml
    outputs:
      artifacts:
      - name: text
        path: /tmp/some-text.txt
```

becoming this:
```yaml
    outputs:
      artifacts:
      - name: text
        path: /tmp/some-text.txt
        # Workflows will tarball all artifacts by default
        # Adding this line prevents it
        archive:
          none: { }
```

If so, very subjective, so will see if other people think differently, but for me the code comment:
```yaml
        # Workflows will tarball all artifacts by default
        # Adding this line prevents it
```

is enough to prompt me to notice the diff, and the fact that the `#vvvvv` is not in the exact place where the new code was introduced threw me off.

_But_, I was looking at the raw markdown code rather than the rendered markdown, in which it's perhaps more obvious what is intended.

docs/how-tos/create-artifacts.md
```diff
@@ -0,0 +1,188 @@
+# Creating artifacts
```
daurer, line 1:
```suggestion
# Creating Artifacts
```

docs/how-tos/create-artifacts.md
```diff
@@ -0,0 +1,188 @@
...
+
+## Preface
+
+This guide will describe how to create artifacts and export them at the end of a workflow run. These artifacts are then made available on the dashboard where they can be downloaded.
```
daurer, line 5:
```suggestion
This guide will describe how to create artifacts and export them at the end of a workflow run. These artifacts are then made available on the [dashboard ](https://workflows.diamond.ac.uk/workflows) where they can be downloaded.
```

docs/how-tos/create-artifacts.md
```diff
@@ -0,0 +1,188 @@
...
+          parameters:
+          - name: command
+            value: |
+              echo "This is my message!" > /tmp/some-text.txt
+              cat /tmp/some-text.txt
+```
+
+## Modifying the Template
+
+We must modify the workflow template to expect to create an artifact. Add the `output` tag with the name and path of the expected artifact.
```
daurer, line 58:
```suggestion
We must modify the workflow template for it to create an artifact. Add the `output` tag with the name and path of the expected artifact.
```

docs/how-tos/create-artifacts.md
```diff
@@ -0,0 +1,188 @@
...
+    outputs:
+      artifacts:
+      - name: text
+        path: /tmp/some-text.txt
+        archive:
+          none: { }
+```
+
+By default, all artifact are tarballs. This can be disabled by adding `none` to the archive, meaning artifact are saved as their raw type. Supported artifact types will be available for preview on the dashboard, but taraballs will have to be downloaded to view their content so we suggest not archiving small artifacts - such as images and text files.
+The complete workflow is provided below for reference.
```
daurer, line 81: Could use "??? example" here to collapse the full manifest ?

daurer (changes requested):

daurer (approved):

PR #525, opened 2025-04-10
configure tracing and ingress for telemetry

iamvigneshwars:
After discussing with the cloud team, agreed to use the centralized tracing. 

The cloud team now have a centralized metrics/tracing solution agreed by Garry before he left. So we connect our otel collector to the centralized otel collector which is still under development, but should be done in a week. 

The traces can be viewed at [https://tracing.diamond.ac.uk/](https://tracing.diamond.ac.uk)


MaxHerbs (approved):
LGTM

PR #538, opened 2025-04-15
graph-proxy s3 client to fetch artifacts

iamvigneshwars:
Instead of relying on the argo-server to fetch the object from the s3 bucket, this s3 client directly creates a downloadable pre-signed url valid for 1 hour.


backend/graph-proxy/src/s3client.rs
```diff
@@ -0,0 +1,57 @@
...
+        config_builder.set_credentials_provider(Some(credentials_provider));
+        config_builder.set_endpoint_url(args.s3_endpoint_url.map(String::from));
+        config_builder.set_force_path_style(Some(args.s3_force_path_style));
+        config_builder.set_region(Some(Region::new(
+            args.s3_region.unwrap_or(String::from("undefined")),
+        )));
+        let config = config_builder.build();
+        Client::from_conf(config)
+    }
+}
```
yousefmoazzam, line 57: Seeing as this involves essentially a conversion from `S3ClientArgs` to `aws_sdk_s3::Client`, and the addition of the `FromS3ClientArgs` trait feeling a tad clunky, I was taking a look at [when `From` is appropriate to be implemented](https://doc.rust-lang.org/std/convert/trait.From.html#when-to-implement-from).

All errors are handled in the logic (so the conversion is infallible), and there seems only way to go from a `S3ClientArgs` to an `aws_sdk_s3::Client`.

If that's correct, I'd be _tempted_ to do something like:
```suggestion

impl From<S3ClientArgs> for Client {
    fn from(value: S3ClientArgs) -> Self {
        let credentials = Credentials::new(
            value.s3_access_key_id.unwrap_or_default(),
            value.s3_secret_access_key.unwrap_or_default(),
            None,
            None,
            "Other",
        );
        let credentials_provider = SharedCredentialsProvider::new(credentials);
        let mut config_builder = aws_sdk_s3::config::Builder::new();
        config_builder.set_credentials_provider(Some(credentials_provider));
        config_builder.set_endpoint_url(value.s3_endpoint_url.map(String::from));
        config_builder.set_force_path_style(Some(value.s3_force_path_style));
        config_builder.set_region(Some(Region::new(
            value.s3_region.unwrap_or(String::from("undefined")),
        )));
        let config = config_builder.build();
        Client::from_conf(config)
    }
}
```

which seems to work for me locally after making a few relevant changes, and it avoids the extra trait `FromS3ClientArgs` (which is essentially `From<S3ClientArgs>` as far as I can tell, but do correct me if that's wrong).

What do you think?

yousefmoazzam (commented):
One question around the newly introduced trait `FromS3ClientArgs` vs. implementing `From<S3ClientArgs>`, would be interested to hear what you think.

backend/graph-proxy/src/s3client.rs
iamvigneshwars: Your are absolutely correct, as we discussed before, I was trying to write some trait for the pre-signed thing to test it, was completely blinded by it 🫤

yousefmoazzam (approved):
Nice, I like that `Task` and `Artifact` have less data attached to them now, looks good!

PR #543, opened 2025-04-16
fix(certificates): new tls certificates

MaxHerbs:
I havent tested the changes to the workflow cluster charts as this cant be done without affecting the vcluster and arent managed by argocd.


daurer (approved):
LGTM

PR #544, opened 2025-04-17
feat(graph-proxy): increase the limit on requested workflows

MaxHerbs:
30 seemed reasonable, does anyone have any input? Maybe 40/50?


daurer (approved):
LGTM, 30 sounds reasonable. Maybe @VictoriaBeilsten-Edmands might have some thoughts?

VictoriaBeilsten-Edmands (approved):
We already discussed. I agree that 30 is good.

PR #547, opened 2025-04-23
feat(dashboard): set number of workflows to list

MaxHerbs:


frontend/relay-workflows-lib/lib/components/Workflows.tsx
```diff
@@ -116,13 +122,32 @@ export default function Workflows({ visit }: { visit: Visit }) {
...
         />
+
+        <FormControl sx={{ width: 80 }}>
+            <InputLabel id="setLimit"># Results</InputLabel>
+            <Select
+              size="small"
+              labelId="setLimitSelector"
+              value={selectedLimit.toString()}
+              label="Results Per Page"
+              onChange={limitChanged}
```
VictoriaBeilsten-Edmands, line 143: Can call `setSelectedLimit(Number(event.target.value))` inline as an arrow function here rather than defining limitChanged.

frontend/relay-workflows-lib/lib/components/Workflows.tsx
```diff
@@ -116,13 +122,32 @@ export default function Workflows({ visit }: { visit: Visit }) {
...
           count={totalPages}
           page={currentPage}
           onChange={handlePageChange}
           showFirstButton
+          siblingCount={1}
+          boundaryCount={0}
         />
+
+        <FormControl sx={{ width: 80 }}>
+            <InputLabel id="setLimit"># Results</InputLabel>
```
VictoriaBeilsten-Edmands, line 137: Maybe `Results per page` for label.

VictoriaBeilsten-Edmands (changes requested):
This doesn't quite work correctly as the cursor history is not changed when changing number of entried per page. 
For example if you are on page 2 with a limit of 10 per page you see entries 11-20. If you change entries per page to 5, you see entries 11-16. If you then go to page 1, you will see entries 1-6, then navigating back to page 2 will jump back to entries 11-16.

frontend/relay-workflows-lib/lib/components/Workflows.tsx
```diff
@@ -116,13 +125,32 @@ export default function Workflows({ visit }: { visit: Visit }) {
...
           count={totalPages}
           page={currentPage}
           onChange={handlePageChange}
           showFirstButton
+          siblingCount={1}
+          boundaryCount={0}
         />
+
+        <FormControl sx={{ width: 80 }}>
+            <InputLabel id="setLimit">Results Per Page</InputLabel>
```
VictoriaBeilsten-Edmands, line 140: The title is too wide for the box. Either make the box wider or you could just put the old title back.

VictoriaBeilsten-Edmands (approved):
I agree that going back to page 1 each time works well.

frontend/relay-workflows-lib/lib/components/Workflows.tsx
MaxHerbs: I have mixed feeling on the old title as well -  what do you think about having no title at all?

VictoriaBeilsten-Edmands: You'd have to add an aria label to make it accessible without the InputLabel.

VictoriaBeilsten-Edmands: It should be fairly obvious with it being next to the pagination.

MaxHerbs: Good - happy to merge if you can reapprove

VictoriaBeilsten-Edmands (approved):
LGTM

PR #553, opened 2025-04-24
Workflow UI templates

TBThomas56:
Added templates page without pagination for now

PR #556, opened 2025-04-25
Workflow UI nav

TBThomas56:


TBThomas56:
Needs to be merged after #553

VictoriaBeilsten-Edmands (approved):
LGTM

VictoriaBeilsten-Edmands:
Shall we move the links to the left in the Navbar, so they're to the right of the Diamond logo, rather than being vaguely centred?
I'm planning to make some changes to the underlying Navbar component to make the positioning easier.

TBThomas56:
> Shall we move the links to the left in the Navbar, so they're to the right of the Diamond logo, rather than being vaguely centred?

Yes I do think that could be could. Is there a reason why the diamond logo has such a large indent on the ui bar. Also maybe that should link to the diamond page?

PR #557, opened 2025-04-25
feat(frontend): add workflow submission page

VictoriaBeilsten-Edmands:


frontend/relay-workflows-lib/lib/components/TemplateView.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+const TemplateViewQuery = graphql`
+  query TemplateViewQuery($templateName: String!) {
+    workflowTemplate(name: $templateName) {
+      ...workflowTemplateFragment
+    }
+  }
+`;
+
+const TemplateViewMutation = graphql`
```
TBThomas56, line 16: More of  a question, why does this not need to be separated as a relay fragment?

VictoriaBeilsten-Edmands: It could be made into a fragment if there were child components that would call it and you would want just that child to update separately from other children. Here, I don't think it's needed (yet).

TBThomas56 (approved):
LGTM

PR #558, opened 2025-04-25
feat(dashboard): added navigation to header

MaxHerbs:
I've added a basic nav bar to the header,
Let me know if this is what you had in mind. I've dropped the "title" section in favor of this - Its still a parameter in the nav bar function but I will remove all mentions of it but I will remove the references to it if  you're happy with this.

We probably need to have a slightly different approach to the formatting of the session part of the header as this moves the title, depending on which pages do/don't have it

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@ kubeconfig.yaml
 
 # Relay
 __generated__
+supergraph.graphql
 
 # File Types
 **/*.tgz # Chart files created when deploying helm dep build
--- a/frontend/dashboard/src/routes/Dashboard.tsx
+++ b/frontend/dashboard/src/routes/Dashboard.tsx
@@ -117,9 +117,7 @@ const argoCdCard = (
 function Dashboard() {
   return (
     <ThemeProvider theme={DiamondTheme} defaultMode="light">
-      <WorkflowsNavbar
-        title={"Workflows at Diamond"}
-      ></WorkflowsNavbar>
+      <WorkflowsNavbar />
       <Container maxWidth="sm" sx={{ mt: 5, mb: 4 }}>
         <Grid
           container
--- a/frontend/dashboard/src/routes/TemplatesList.tsx
+++ b/frontend/dashboard/src/routes/TemplatesList.tsx
@@ -7,7 +7,7 @@ const TemplatesList: React.FC = () => {
   return (
     <>
       <ThemeProvider theme={DiamondTheme} defaultMode="light">
-        <WorkflowsNavbar title="Templates" />
+        <WorkflowsNavbar />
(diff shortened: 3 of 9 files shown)


VictoriaBeilsten-Edmands:
I agree that no title on the page would be better. I removed it on one of the pages I was working on as I think it's better without. Having the Navlinks works well. An addition we could have too would be to change the new back button in #556 to [Breadcrumbs](https://diamondlightsource.github.io/sci-react-ui/?path=/docs/scireactui-navigation-breadcrumbs--docs) and put it in the same place (top left below Navbar). What do you think? @TBThomas56

VictoriaBeilsten-Edmands:
I suggest merging this in. We should lose the titles on the page in favour of these links.

MaxHerbs:
I've fixed quite a few merge conflicts there so if you wouldn't mind having another skim over that thanks.

frontend/relay-workflows-lib/supergraph.graphql
VictoriaBeilsten-Edmands, line 1: This gets added by CI

frontend/dashboard/src/routes/Dashboard.tsx
```diff
@@ -118,7 +118,7 @@ function Dashboard() {
   return (
     <ThemeProvider theme={DiamondTheme} defaultMode="light">
       <WorkflowsNavbar
-        title={"Workflows at Diamond"}
+        sessionInfo={""}
```
VictoriaBeilsten-Edmands, line 121: Can you remove this or does it cause other elements to move?

MaxHerbs: Looks like I reintroduced that when I was fixing a merge conflict, I'll remove it. I think it might be worth moving the session info somewhere else on the page in future though as it does push the menu text around - meaning the navbar text is in a different place on the index, the search with no session selected, and the search with a session selected. I'm also aware theres a diamond default "profile" widget that goes in that place.

VictoriaBeilsten-Edmands: I'm planning to update the Navbar component to make it easier to choose where components go to prevent them getting pushed along.

MaxHerbs: Okay perfect not a worry then

.gitignore
```diff
@@ -9,6 +9,7 @@ kubeconfig.yaml
 
 # Relay
 __generated__
+supergraph.graphql
```
iamvigneshwars, line 12: We should avoid dev struff to the `main` even in `.gitignore`

iamvigneshwars (changes requested):

iamvigneshwars (approved):

PR #570, opened 2025-05-02
Filter group backend

MaxHerbs:


backend/graph-proxy/.devcontainer/devcontainer.json
```diff
@@ -6,7 +6,8 @@
   },
   "containerEnv": {
     "ARGO_SERVER_URL": "https://argo-workflows.workflows.diamond.ac.uk/",
-    "ARGO_SERVER_SCHEMA_URL": "https://argo-workflows.workflows.diamond.ac.uk/assets/openapi-spec/swagger.json"
+    "ARGO_SERVER_SCHEMA_URL": "https://argo-workflows.workflows.diamond.ac.uk/assets/openapi-spec/swagger.json",
+    "S3_BUCKET": "test-bucket"
```
MaxHerbs, line 10: @iamvigneshwars I've put this in to appease the CLI parer so graph-proxy runs, buts its occurred to me this wont actually allow testing/dev of the s3 functions. Is this a non-issue, or do we need to add an s3 mock to the dev-container (or something along those lines)? I assume you were using workflows test-bucket when you were working on this?

iamvigneshwars: We need the s3 bucket to query the artifacts, so that can't be optional
Testing don't relay on the actual bucket, so it should be fine.
We are using the actual argo server for the dev, so if  we add a s3 mock to the dev-container, it will be useless as it will be empty.  For dev, I pointed it to the actual s3 bucket that argo server use.
I think we can point it to the actual s3 bucket in the `devcontainer.json` like `S3_BUCKET: k8s-workflows-test` and `S3_ENDPOINT_URL: https://sci-nas-s3.diamond.ac.uk` and put `access-key` and `secret-key` in a `.env`.
what do you think?

iamvigneshwars (approved):
LGTM

PR #575, opened 2025-05-06
add Thomas to argocd admins

MaxHerbs:


charts/argocd/values.yaml
```diff
@@ -29,7 +29,7 @@ argo-cd:
         g, benedikt.daurer@diamond.ac.uk, role:admin
         g, garry.o'donnell@diamond.ac.uk, role:admin
```
daurer, line 30: Might as well remove Garry from admins

daurer (changes requested):

daurer (approved):

PR #590, opened 2025-05-16
Restrict UI parameters

MaxHerbs:
Dont show parameters from configmap when no Json form is provided
The API reference for `valueFrom` is [here](https://argo-workflows.readthedocs.io/en/latest/fields/#valuefrom), I dont see why any of these should appear in  the UI but we can revisit if a usecase surfaces


iamvigneshwars (approved):
LGTM

PR #607, opened 2025-05-27
feat(frontend): report errors/success on template submission

VictoriaBeilsten-Edmands:
On template submission, populates an accordion list with either error info or successfully submitted workflow name (with link to workflow page).

I put the submission message types in the workflows lib even though they have a relay import (for PayloadError). I can move this into the relay lib if we think it fits there better.

### DIFF
--- a/frontend/workflows-lib/lib/components/workflow/SubmittedMessagesList.tsx
+++ b/frontend/workflows-lib/lib/components/workflow/SubmittedMessagesList.tsx
@@ -1,0 +1,126 @@
+import React from "react";
+import {
+  Accordion,
+  AccordionSummary,
+  AccordionDetails,
+  Box,
+  Divider,
+  Paper,
+  Typography,
+} from "@mui/material";
+import ExpandMoreIcon from "@mui/icons-material/ExpandMore";
+import { Link } from "react-router-dom";
+import {
+  SubmissionGraphQLErrorMessage,
+  SubmissionNetworkErrorMessage,
+  SubmissionSuccessMessage,
+} from "../../types";
+
+const renderSubmittedMessage = (
+  r:
+    | SubmissionGraphQLErrorMessage
+    | SubmissionNetworkErrorMessage
+    | SubmissionSuccessMessage,
+  i: number,
+) => {
+  switch (r.type) {
+    case "success":
+      return (
+        <Accordion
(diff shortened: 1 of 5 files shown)


MaxHerbs:
Looks good, I haven't prompted every error case but works well,
I wouldn't say importing the relay component is an issue, seems like the cleanest solution I can think of

frontend/workflows-lib/lib/types.ts
TBThomas56, line 1: pulling the branch shows trailing whitespace after:
- message:string
- the "}" after that
- error:Error
- errors:PayloadError[]

frontend/workflows-lib/lib/components/workflow/SubmittedMessagesList.tsx
```diff
@@ -0,0 +1,126 @@
...
+                </Typography>
+              );
+            })}
+          </AccordionDetails>
+        </Accordion>
+      );
+  }
+};
+
+interface SubmittedMessagesListProps {
```
TBThomas56, line 85: Any decision why the interface was kept here and not in the types.ts file?

frontend/workflows-lib/lib/components/workflow/SubmittedMessagesList.tsx
```diff
@@ -0,0 +1,126 @@
...
+        maxWidth: "1200px",
+        height: "100%",
+        mx: "auto",
+      }}
+    >
+      <Divider sx={{ mt: 5, mb: 5 }} />
+      {submissionResults.length > 0 && (
+        <>
+          <Paper elevation={1} sx={{ p: 2 }}>
+            <Typography variant="h6" sx={{ mb: 2, textAlign: "center" }}>
```
TBThomas56, line 115: More of a nitpick. the h6 for Submissions looks quite small in the box it's located in. Either maybe bump that up to h5 (still smaller than the template title -h4) or consider reducing the size of the box?

TBThomas56 (changes requested):
Looks really good and functionality works. Don't think any changes to that is needed

frontend/workflows-lib/lib/types.ts
VictoriaBeilsten-Edmands: These should be fixed now. The semicolons weren't right either.

PR #609, opened 2025-05-28
feat(dashboard): added workflow status select component to workflow f…

TBThomas56:
Added on to filtering by creator to get both filters working in tandem. 
Worth noting that the appending of the filter is set verbosely as there were typing errors in checking which status was being used the normaliseFilter function


frontend/workflows-lib/lib/components/workflow/WorkflowListFilterDrawer.tsx
```diff
@@ -37,6 +40,7 @@ function WorkflowListFilterDrawer({ onApplyFilters }: WorkflowListFilterDrawerPr
   const [open, setOpen] = useState(false);
   const [creator, setCreator] = useState<string>("");
   const [errors, setErrors] = useState<{ creator?: boolean }>({});
+  const [status, setStatus] = useState<WorkflowStatusBool | undefined>(undefined);
```
VictoriaBeilsten-Edmands, line 43: Would be better here to initialize to all true rather than undefined. Then once the filters are selected, in the UI only list the filters that are true i.e. instead of displaying "workflowStatusFilter: { "pending": true, "running": false, "succeeded": true, "failed": true, "error": false }", something more like: "Workflow Status Filter: pending, succeeded, failed"

frontend/workflows-lib/lib/components/workflow/WorkflowListFilterDrawer.tsx
```diff
@@ -47,6 +51,21 @@ function WorkflowListFilterDrawer({ onApplyFilters }: WorkflowListFilterDrawerPr
     setErrors((prev) => ({ ...prev, creator: !isValid }));
   };
 
+  const handleStatusChange = (value: string[]) => {
+    if (value.length === 0) {
+      setStatus(undefined);
```
VictoriaBeilsten-Edmands, line 56: clear to all true

frontend/workflows-lib/lib/components/workflow/WorkflowListFilterDrawer.tsx
```diff
@@ -134,6 +145,29 @@ function WorkflowListFilterDrawer({ onApplyFilters }: WorkflowListFilterDrawerPr
...
+                : []
+            }
+            onChange={(e) => { handleStatusChange(e.target.value as string[]); }} 
+            label="Status"
+          >
+            <MenuItem value={"pending"}>Pending</MenuItem>
+            <MenuItem value={"running"}>Running</MenuItem>
+            <MenuItem value={"succeeded"}>Succeeded</MenuItem>
+            <MenuItem value={"failed"}>Failed</MenuItem>
+            <MenuItem value={"error"}>Errored</MenuItem>
```
VictoriaBeilsten-Edmands, line 167: Consistency in naming: error vs errored

VictoriaBeilsten-Edmands (changes requested):

frontend/workflows-lib/lib/components/workflow/WorkflowListFilterDrawer.tsx
MaxHerbs: I had suggested undefined rather than all true so the query wouldn't be used at all - I'm not certain on the database query implications, but querying by status does something like `status in success, running, error` with the allowed statuses, so that would check against all possible statuses (ie all for the initial state) which is a bit redundant for the 'show me everything' case - where as undefined will drop the dropped the status filter altogether so won't be taken into account at all - maybe this is something for the `normaliseFilter` function to handle? 

Also when the filters are defined they'll show on the page which might not be desirable when the user loads the session page - again this is probably fixable in the same way

frontend/workflows-lib/lib/components/workflow/WorkflowListFilterDrawer.tsx
TBThomas56: the value was set as error to match the value found in the backend and thought i keep consistency with the frontend. Do you think it better to have both as Error then?

frontend/workflows-lib/lib/components/workflow/WorkflowListFilterDrawer.tsx
```diff
@@ -22,7 +36,16 @@ export const WorkflowListFilterDisplay = ({ filter }: { filter?: WorkflowQueryFi
           .filter(([, val]) => val)
           .map(([key, value], index) => (
             <div key={index}>
```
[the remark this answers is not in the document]
TBThomas56: Fixed

VictoriaBeilsten-Edmands (approved):
This looks good and the filtering works well.

PR #613, opened 2025-05-30
feat(frontend): add link to single workflow from accordion

VictoriaBeilsten-Edmands:
Added link to single workflow view to accordion on workflow list page but not single workflow page.


MaxHerbs (approved):
LGTM

PR #617, opened 2025-06-02
feat(frontend): add rerun template from submitted workflow

VictoriaBeilsten-Edmands:
Adds functionality to rerun workflow from same template and input parameters. Links from workflow page to `template page/visit-workflowName`, prepopulating with input params. I've had to add a couple of layers to prevent conditionally calling hooks etc.


MaxHerbs (approved):
Looks good, the extra route to get to the template without the overridden parameters is a nice addition!

PR #621, opened 2025-06-05
feat(frontend): return all templates without limit

VictoriaBeilsten-Edmands:


frontend/relay-workflows-lib/lib/components/TemplatesList.tsx
```diff
@@ -25,7 +19,6 @@ const templatesListQuery = graphql`
 
 export default function TemplatesList() {
   const data = useLazyLoadQuery<TemplatesListQuery>(templatesListQuery, {
-    limit: 10,
     cursor: null,
```
TBThomas56, line 22: Do we need the cursor or can that be removed as well?

TBThomas56 (approved):
LGTM

PR #632, opened 2025-06-09
Staging

MaxHerbs:
This is a quite a big PR, but probably needs to all be merged in one piece, and may well have teething problems in Argus when we merge it. Happy to take any input around what the best process is to merge it. I'm going to try and slim down the number of commits a bit


MaxHerbs:
That works for me - alternatively I can shard it into a few smaller PR and release them in the appropriate order? I couldn't do this orignally as sessionspaces was blocking me, but I should be able to trim down a little

PR #633, opened 2025-06-10
feat(graph-proxy): add id to Workflow to allow refetchable fragments

VictoriaBeilsten-Edmands:
Duplicates name


MaxHerbs:
`cargo fmt` should sort this

PR #639, opened 2025-06-12
fix(frontend): use modified visitRegex in template regex

VictoriaBeilsten-Edmands:
Use visitRegex without end of line for template regex


TBThomas56 (approved):
LGTM

PR #648, opened 2025-06-19
Workflows CLI

MaxHerbs:


workflows-cli/src/linter/helm.rs
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    write_to_clean_folder(tmp_dir, manifests)
+        .map_err(|_e| "Couldn't create temporary file for helm templates.")?;
+
+    let path_buf = tmp_dir.to_path_buf();
+    lint_from_manifest(&path_buf, true)
+}
+
+pub fn helm_template(target: &Path, template: Option<&str>) -> Result<Vec<String>, String> {
```
davehadley, line 49: I think that all functions in this file except lint_from_helm can be private

workflows-cli/src/linter/helm.rs
```diff
@@ -0,0 +1,120 @@
...
+use std::fs::{File, read_dir};
+use std::io::Write;
+use std::path::Path;
+
+use crate::command_runner::get_command_factory;
+use crate::linter::LintResult;
+use crate::linter::base_linting::lint_from_manifest;
+
+pub fn lint_from_helm(target: &Path, all: bool) -> Result<Vec<LintResult>, String> {
+    let tmp_dir = Path::new("/tmp/argo-lint");
```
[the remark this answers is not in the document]
davehadley: It might be easier and safer to use a "tempfile" / "tempdir" crate to handle these temp directory shenanigans for you.

workflows-cli/src/command_runner.rs
```diff
@@ -0,0 +1,143 @@
...
+    /// Initiate mock command runner
+    pub fn new(cmd: &str) -> Self {
+        let yaml_str = fs::read_to_string("./tests/mock_commands.yaml")
+            .expect("Failed to read mock_commands.yaml");
+
+        let mappings: HashMap<String, Vec<MockEntry>> =
+            serde_yaml::from_str(&yaml_str).expect("Failed to parse YAML");
+
+        let target_mapping = env::var("RUST_TEST_ACTIVE_MAPPING").unwrap();
+        let active_mapping = mappings.get(&target_mapping).unwrap();
```
iamvigneshwars, line 84: Is this safe? could we return `Self` as a `Result` ?

workflows-cli/src/linter/mod.rs
```diff
@@ -0,0 +1,212 @@
...
+    fn have_same_elements<T: Ord>(result: &mut Vec<T>, expected: &mut Vec<T>) -> bool {
+        result.sort();
+        expected.sort();
+        result == expected
+    }
+
+    #[test]
+    #[serial]
+    fn lint_single_passing_manifest() {
+        unsafe {
```
davehadley, line 86: Unsafe usage should usually be annotated with a safety comment.

workflows-cli/src/main.rs
```diff
@@ -0,0 +1,89 @@
...
+        Cli::Lint(args) => {
+            linter::lint(args);
+        }
+        Cli::LintConfig(args) => {
+            config_lint(args);
+        }
+    }
+}
+
+use std::process::Command;
```
iamvigneshwars, line 66: could we move this to the top?

workflows-cli/tests/mock_commands.yaml
iamvigneshwars, line 1: EOF

workflows-cli/src/command_runner.rs
```diff
@@ -0,0 +1,143 @@
...
+
+        Self {
+            cmd: cmd.to_string(),
+            args: vec![],
+            mappings: active_mapping.clone(),
+        }
+    }
+
+    /// Lookup command in the mapping table
+    fn get_response(&self) -> (String, i32) {
```
iamvigneshwars, line 94: similarly could this be an `Option`?

workflows-cli/src/linter/helm.rs
iamvigneshwars, line 13: this could sometimes cause race conditions if multiple instances are running, we might have to use `tempfile` crate here

workflows-cli/src/linter/mod.rs
```diff
@@ -0,0 +1,212 @@
...
+            env::set_var("RUST_TEST_ACTIVE_MAPPING", "passing_manifest");
```
davehadley, line 87: Why did you choose to modify the behavior of the MockCommand via environment variables? This seems to me that it might be potentially confusing to a future maintainer.

As a minimum, I would suggest adding comments to explain to a future maintainer what these environment variables are doing.

workflows-cli/src/linter/helm.rs
```diff
@@ -0,0 +1,120 @@
...
+
+    let target = match target.is_absolute() {
+        true => target,
+        false => &Path::new("./").join(target),
+    };
+
+    let select_template = match all {
+        true => None,
+        false => Some(target.file_name().unwrap().to_str().unwrap()),
```
iamvigneshwars, line 22: is this safe?

workflows-cli/src/command_runner.rs
```diff
@@ -0,0 +1,143 @@
...
+impl MockCommand {
+    /// Initiate mock command runner
+    pub fn new(cmd: &str) -> Self {
+        let yaml_str = fs::read_to_string("./tests/mock_commands.yaml")
+            .expect("Failed to read mock_commands.yaml");
+
+        let mappings: HashMap<String, Vec<MockEntry>> =
+            serde_yaml::from_str(&yaml_str).expect("Failed to parse YAML");
+
+        let target_mapping = env::var("RUST_TEST_ACTIVE_MAPPING").unwrap();
```
davehadley, line 83: It would be best to "namespace" any special environment variables that our application depends on.

eg WORKFLOWS_CLI_TEST_ACTIVE_MAPPING or similar

workflows-cli/src/command_runner.rs
```diff
@@ -0,0 +1,143 @@
...
+/// New command trait fo the mock factory
+impl CommandFactory for MockCommandFactory {
+    fn new_command(&self, cmd: &str) -> Box<dyn CommandLike> {
+        Box::new(MockCommand::new(cmd))
+    }
+}
+
+/// Get an instance of the CLI builder
+pub fn get_command_factory() -> Box<dyn CommandFactory> {
+    match env::var("MOCK_COMMAND") {
```
davehadley, line 139: It would be best to "namespace" any special environment variables that our application depends on.

eg WORKFLOWS_CLI_TEST_ENABLE_MOCK_COMMAND or similar

davehadley: Why did you choose to have the mocking live in the real application and not exist only in the test module?

Is the intention to have the ability to run the tests using the real "argo" and "helm" commands if they are installed on the system?

workflows-cli/src/linter/config_linting.rs
```diff
@@ -0,0 +1,61 @@
...
+                Err(e) => {
+                    eprintln!("Error linting chart '{}': {}", chart.display(), e);
+                    std::process::exit(1);
+                }
+            }
+        }
+    }
+
+    print_and_exit(lint_results);
+}
```
iamvigneshwars, line 54: could we decouple this logic from process termination? we could propogate the error back and process termination could happen at the top of the stack?

workflows-cli/src/linter/mod.rs
```diff
@@ -0,0 +1,212 @@
...
+        } else {
+            println!("Errors found");
+            result.errors.iter().for_each(|e| println!("  - {}", e));
+        }
+    });
+    match results.iter().any(|result| !result.errors.is_empty()) {
+        true => std::process::exit(1),
+        false => std::process::exit(0),
+    }
+}
```
iamvigneshwars, line 66: could we decouple this logic from process termination? we could propogate the error back and process termination could happen at the top of the stack?

workflows-cli/src/linter/helm.rs
MaxHerbs: Yes this is a good catch. I'll have a look into that

.github/workflows/_cli_code.yaml
```diff
@@ -0,0 +1,125 @@
...
+    needs: [test, lint]
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Install stable toolchain
+        uses: actions-rust-lang/setup-rust-toolchain@v1.12.0
+
+      - name: Download tarpaulin
+        run: |
+          curl -L https://github.com/xd009642/tarpaulin/releases/latest/download/cargo-tarpaulin-x86_64-unknown-linux-gnu.tar.gz | tar xz
```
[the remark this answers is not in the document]
MaxHerbs: Installing with cargo builds the tarpaulin binary which can take ~2 minutes in CI. Tarpaulin then rebuilds everything so it can to do code coverage, which can take another 2/3 minutes. Downloading the binary instead is much quicker, and takes a second or two instead of minutes, which generally makes the processes faster.

workflows-cli/src/linter/helm.rs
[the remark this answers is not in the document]
MaxHerbs: I'll have a look at some crates instead, that's a good idea

workflows-cli/src/linter/helm.rs
```diff
@@ -0,0 +1,120 @@
...
+use crate::command_runner::get_command_factory;
+use crate::linter::LintResult;
+use crate::linter::base_linting::lint_from_manifest;
+
+pub fn lint_from_helm(target: &Path, all: bool) -> Result<Vec<LintResult>, String> {
+    let tmp_dir = Path::new("/tmp/argo-lint");
+
+    let target = match target.is_absolute() {
+        true => target,
+        false => &Path::new("./").join(target),
```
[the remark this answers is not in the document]
MaxHerbs: I've added the following comment:

```rust
    // When --all is provided, we traverse up the file path to find the directory with the chart.yaml
    // in it. If a user passes 'templates/workflow.yaml', this will check only the templates dir,
    // and doesnt check '.' as this is not explicitly in the path which will cause the lint to fail
    // if lint is called from the same directory as the Chart.yaml.
    // Appending './' adds '.' to the path traversal, which makes sure the working directory is also
    // checked for Chart.yaml
```

.github/workflows/_cli_code.yaml
```diff
@@ -0,0 +1,125 @@
...
+      - name: Run coverage
+        run: |
+          cd workflows-cli
+          MOCK_COMMAND=1 cargo tarpaulin --out html
+
+      - name: Upload coverage report as artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: coverage.html
+          path: workflows-cli/tarpaulin-report.html
```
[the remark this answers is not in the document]
MaxHerbs: It seems to work really badly with workspaces - I gave it a go, but it give one coverage for all of them, and includes the 80,000 lines of workflows structs that are created from the swagger schema so it comes out to like 1%. I only had a quick look so there might be config somewhere in it to handle that better

workflows-cli/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "workflows"
```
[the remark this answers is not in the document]
MaxHerbs: It would name the binary workflows-cli in that case - possibly doesnt matter? I've overwritten the package name in release please to workflows-cli for our clarity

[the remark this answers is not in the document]
iamvigneshwars: Ahhh got it

MaxHerbs:
This should be ready for re-review. I've implimented most of the requested changes except;

I've left the tmpdir as is. Moving to a tempdir module was pretty easy, but this broke the tests as the path wasn't determinate anymore so I reverted that. I can spend a bit more time investigating that, but I dont see many cases where the race condition would surface as this will usually be run on personal workspaces, and the chances of a clash are pretty low.

More generally, I've added a more comments especially around the testing to clarify how that works for anyone working on this in future.

iamvigneshwars (approved):
LGTM

davehadley (approved):

PR #662, opened 2025-06-23
ci: better caching for graph proxy

MaxHerbs:
I think we should move our caching for graph proxy/sessionspaces to this rather than the swatinem action. Swatinem seems to average around 1:40 for the main compile portion of the CI, vs 20-30 seconds for this caching. Happy to investigate further. Probably beneficial to frontend as well

### DIFF
--- a/.github/workflows/_graph_proxy_code.yaml
+++ b/.github/workflows/_graph_proxy_code.yaml
@@ -18,11 +18,10 @@ jobs:
           cache: false
           components: clippy,rustfmt
 
-      - name: Cache Rust Build
-        uses: Swatinem/rust-cache@v2.7.8
+      - uses: actions/cache@v4.2.3
         with:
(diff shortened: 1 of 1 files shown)


davehadley:
Do you understand what the difference is?

It looks like Swatinem/rust-cache (https://github.com/Swatinem/rust-cache) only caches dependencies and disables incremental compilation. So maybe that is the explanation? There is an active pull request to implement this https://github.com/Swatinem/rust-cache/pull/227. Reading through it seems that this is a complicated topic... 

Are you sure that you want to go ahead with this change? Maintaining our own solution may be more costly that the extra seconds of CI time. It might be easier to wait for the community action performance to improve.

Also, it looks like the existing github action also caches ~/.cargo, so you may want to do that as well, if you want to go ahead with this change?

PR #669, opened 2025-06-25
WorkflowTemplates Lint Action

MaxHerbs:
The second step in the examples dag that I've removed fails linting. The issue is related to the volume mount, but this is necessary for the second step to work. The argo CLI is wrong - but the fail is unavoidable so I've taken the whole step out so the CI will run. In future, this fail should be avoided by assigning the volumeMounts to the underlying template rather than when referencing the templates.


.github/workflows/_lint_workflows.yaml
```diff
@@ -0,0 +1,14 @@
...
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Source
+        uses: actions/checkout@v4.2.2
+
+      - name: Lint Workflows
+        uses: diamondlightsource/workflows@action
```
MaxHerbs, line 14: This line is a bit of a pain. This should be main, but the action doesnt exist in main yet. Either we can switch it to main now, and ignore the failing CI, or we can leave it as the 'action' branch, and make a follow up PR to switch it over. I think switching it to main would be cleaner.

davehadley (dismissed):

.github/workflows/_lint_workflows.yaml
MaxHerbs: I think I prefer switching to main - I'll push a new version and that should be ready to merge

examples/conventional-templates/using-cluster-templates.yaml
```diff
@@ -68,11 +68,3 @@ spec:
             value: "{{workflow.parameters.size}}"
           - name: memory
             value: "{{workflow.parameters.memory}}"
-      - name: mount
```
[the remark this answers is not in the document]
MaxHerbs: Yes its to fix the failing linting. The 'volumes' section was added in a recent PR but should really have been fixed in the underlying workflow template not at this level. This is the simplest fix that both the templates and the linting in a working state without making this PR too large

davehadley (approved):

PR #670, opened 2025-06-25
feat(groups): Add new imaging-workflows/nxstacker directory.

davehadley:


MaxHerbs:
Looks good, what was the motivation for making the nxtracker templates helm based rather than regular manifests?

PR #696, opened 2025-07-01
fix(action): checkout code in action

MaxHerbs:


davehadley:
How did this ever work if it wasn't checking out the code before? Was it just completely broken before?

davehadley (approved):

PR #736, opened 2025-07-16
Centralise groups, and give 'get' rbac

MaxHerbs:


charts/groups/templates/apps-rolebinding.yaml
```diff
@@ -0,0 +1,27 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: app-viewr
```
iamvigneshwars, line 4: typo `app-viewer`

iamvigneshwars (changes requested):

iamvigneshwars:
@MaxHerbs I would lean more towards hiding behind the graph for long-term flexibility ,we could have more granular control over access to specific apps without exposing everything. 

But I'm happy with the current approach as we currently don't have any privacy risks.

PR #740, opened 2025-07-16
Shortcuts for task selection

jacob720:
Adds an `OUTPUT` and `CLEAR` button on the single workflow page as shortcuts for selecting output tasks and clearing all selected tasks.


jacob720:
> Assuming there wasn't one in a previous PR that I've missed - could we get a test to check the highlighted task selection logic? Maybe test for specific cases like when the button is clicked so components don't get accidentality broken in future refactors.

Good idea, I'll add this

PR #750, opened 2025-07-22
Highlight tasks

jacob720:
When artifacts are hovered over with the mouse, the task that produced it is highlighted.

### DIFF
--- a/frontend/relay-workflows-lib/lib/components/WorkflowRelay.tsx
+++ b/frontend/relay-workflows-lib/lib/components/WorkflowRelay.tsx
@@ -4,13 +4,12 @@ import "react-resizable/css/styles.css";
 import { Box } from "@mui/material";
 
 import { TasksFlow, WorkflowAccordion } from "workflows-lib";
-import type { Task, TaskStatus, WorkflowStatus } from "workflows-lib";
+import type { WorkflowStatus } from "workflows-lib";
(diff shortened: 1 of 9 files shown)


frontend/relay-workflows-lib/lib/components/WorkflowRelay.tsx
```diff
@@ -199,8 +201,9 @@ const WorkflowRelay: React.FC<WorkflowRelayProps> = ({
           <TasksFlow
             workflowName={workflowName}
             tasks={fetchedTasks}
-            highlightedTaskNames={selectedTasks}
             onNavigate={onNavigate}
+            highlightedTaskNames={selectedTasks}
+            filledTaskName={filledTaskName ? filledTaskName : null}
```
davehadley, line 206: this use of the ternary operator seems to do nothing?

davehadley (approved):

frontend/relay-workflows-lib/lib/components/WorkflowRelay.tsx
jacob720: This should convert to `null` from `undefined` (although my understanding of the difference is murky).

jacob720:
The bug I mentioned was fixed by debouncing getting the highlighted and filled tasks.

PR #757, opened 2025-07-25
fix(groups): change httomo source type to helm

yousefmoazzam:
This PR corresponds with the following PR in the imaging-workflows repo: https://github.com/DiamondLightSource/imaging-workflows/pull/34

### DIFF
--- a/charts/groups/Chart.yaml
+++ b/charts/groups/Chart.yaml
@@ -3,4 +3,4 @@ name: group-projects
 description: A collection of AppProjects and Apps for science domain groups
 type: application
 
-version: 0.3.9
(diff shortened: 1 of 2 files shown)


yousefmoazzam:
Thanks for the approval @davehadley!

I'm going to merge anyway despite the failing frontend tests because I'm pretty sure my changes have nothing to do with that. Someone may want to keep an eye out for that issue if it's persistent, I got it twice in a row here, pinging @TBThomas56 and @MaxHerbs just in case.

davehadley:
The failing front-end test is a known flaky test that it is safe for you to ignore. 👍

PR #761, opened 2025-07-31
feat(frontend): add ScanRangeInput component and json forms renderer

VictoriaBeilsten-Edmands:
Adds ScanRangeInput component which is a scan range input field with start end and optional excluded inputs as text fields. For start and end these validate positive integers (and 0) where end is not less than start. The excluded input validates a string of comma separated integers in the range of start and end. It also trims whitespace and square brackets before converting to a number array. ScanInputRange is wrapped for use as a JSON forms custom renderer with tester function.

### DIFF
--- a/frontend/workflows-lib/lib/components/template/controls/ScanRangeInput.tsx
+++ b/frontend/workflows-lib/lib/components/template/controls/ScanRangeInput.tsx
@@ -1,0 +1,143 @@
+import { useCallback, useEffect, useState } from "react";
+import { Box, InputAdornment, TextField, Tooltip } from "@mui/material";
+import { InfoOutlined } from "@mui/icons-material";
+import { ScanRangeInputProps } from "../../../types";
+import { validateScanRange } from "../../../utils/validationUtils";
+
+const ScanRangeInput = ({ name, value, handleChange }: ScanRangeInputProps) => {
+  const [scanRange, setScanRange] = useState({
+    start: String(value.start),
+    end: String(value.end),
+  });
+
+  const [excludedRaw, setExcludedRaw] = useState(
+    (value.excluded ?? []).join(", "),
+  );
+  const [errors, setErrors] = useState({
+    start: "",
+    end: "",
+    excluded: "",
+  });
+
+  const [hasUserEditedExcluded, setHasUserEditedExcluded] = useState(false);
+
+  const validateAndUpdate = useCallback(() => {
+    const result = validateScanRange(
+      scanRange.start,
+      scanRange.end,
+      excludedRaw,
+    );
+    setErrors(result.errors);
+
+    if (result.parsed) {
+      const current = {
+        start: Number(scanRange.start),
+        end: Number(scanRange.end),
+        excluded: result.parsed.excluded,
+      };
+
+      const isEqual =
+        current.start === value.start &&
+        current.end === value.end &&
+        JSON.stringify(current.excluded) ===
+          JSON.stringify(value.excluded ?? []);
+
+      if (!isEqual) {
+        handleChange(name, result.parsed);
+      }
+    }
+  }, [scanRange, excludedRaw, handleChange, name, value]);
+
+  const handleFieldChange =
+    (field: keyof typeof scanRange) =>
+    (event: React.ChangeEvent<HTMLInputElement>) => {
+      setScanRange({ ...scanRange, [field]: event.target.value });
+    };
+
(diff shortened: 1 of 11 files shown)


frontend/workflows-lib/lib/utils/validationUtils.ts
```diff
@@ -0,0 +1,102 @@
+import { ScanRange } from "../types";
+
+export const isStrictPositiveInteger = (value: string): boolean =>
+  /^\d+$/.test(value) && Number(value) >= 0;
+
+export const parseExcludedScans = (input: string): number[] => {
+  const cleaned = input.replace(/[[\]]/g, "");
+
+  const parts = cleaned
+    .split(",")
```
davehadley, line 10: I think that this requires input in the form "1, 2, 3" etc . Do we want to be more permissive and accept other separators (eg any whitespace)?

frontend/workflows-lib/lib/utils/validationUtils.ts
```diff
@@ -0,0 +1,102 @@
...
+      const start = Number(startStr);
+      const end = Number(endStr);
+      if (start >= end) {
+        throw new Error("Invalid range");
+      }
+      for (let i = start; i <= end; i++) {
+        resultSet.add(i);
+      }
+    } else {
+      throw new Error("Values must be positive integers");
```
davehadley, line 36: To get to this point what the user could have entered anything and maybe they don't understand the syntax that they are supposed to use.

It may be better to explicitly specify the format that this input expects (currently a comma-separated list of positive integers).

frontend/workflows-lib/lib/components/template/controls/ScanRangeInput.tsx
```diff
@@ -0,0 +1,114 @@
...
+          (event.target as HTMLInputElement).blur();
+        }}
+      />
+      <TextField
+        label="Excluded"
+        value={excludedRaw}
+        onChange={handleExcludedChange}
+        onBlur={validateAndUpdate}
+        error={!!errors.excluded}
+        helperText={errors.excluded || " "}
```
davehadley, line 104: I'm not sure how this looks but did you consider setting the default helper text in the absence of errors should tell the user the syntax expected for excluded runs so that they immediately know how to use this input?

frontend/workflows-lib/lib/utils/validationUtils.ts
```diff
@@ -0,0 +1,102 @@
+import { ScanRange } from "../types";
+
+export const isStrictPositiveInteger = (value: string): boolean =>
+  /^\d+$/.test(value) && Number(value) >= 0;
+
+export const parseExcludedScans = (input: string): number[] => {
+  const cleaned = input.replace(/[[\]]/g, "");
```
davehadley, line 7: Why do we accept square brackets but then just strip them out? This seems to add no utility for the user?

davehadley (approved):

TBThomas56 (dismissed):
LGTM

frontend/workflows-lib/lib/components/template/controls/ScanRangeInput.tsx
VictoriaBeilsten-Edmands: I've added a tooltip that can be displayed on hover with some explanation.

frontend/workflows-lib/lib/utils/validationUtils.ts
VictoriaBeilsten-Edmands: This is to make it compatible with what some beamlines already expect when entering these kind of values.

frontend/workflows-lib/lib/components/template/controls/ScanRangeInput.tsx
```diff
@@ -0,0 +1,114 @@
...
+        alignItems: "flex-start",
+        paddingTop: "20px",
+      }}
+    >
+      <TextField
+        type="number"
+        label="Start"
+        value={scanRange.start}
+        onChange={handleFieldChange("start")}
+        onBlur={validateAndUpdate}
```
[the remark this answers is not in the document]
VictoriaBeilsten-Edmands: I've updated this to check and validate with onChange. It's not entirely straightforward because of the need for cross-field validation and allow exclude to be empty. It works in my testing.

frontend/workflows-lib/lib/utils/validationUtils.ts
VictoriaBeilsten-Edmands: I've added whitespace separation.

PR #783, opened 2025-08-06
feat(frontend): added table sorting and its tests

TBThomas56:


frontend/workflows-lib/tests/components/ArtifactFilteredList.test.tsx
```diff
@@ -4,38 +4,35 @@ import "@testing-library/jest-dom";
 import { mockArtifacts } from "./data";
 
 describe("ArtifactFilteredList", () => {
-  it("renders all artifacts by default", () => {
```
MaxHerbs, line 7: Are these tests redundant now?

TBThomas56: yes they are since they were to test the functionality fo the togglegroup button that was replaced by the search bar and the filetype dropdown

PR #805, opened 2025-08-11
feat(frontend): allow all image types

jacob720:
Fixes https://jira.diamond.ac.uk/browse/AP-817
Counts any artifact with a mime type that begins with `image/` as an image.


jacob720:
> This looks good. Do you have an example workflow with different image types to show this?

The only one I could find had a tiff image, which the browser couldn't display. Would be good to add one with a jpeg.

PR #809, opened 2025-08-11
feat(frontend): store instrumentSessionID in localStorage

VictoriaBeilsten-Edmands:
Removes initial visitID input page and redirects to instrumentSessionID in localStorage.


jacob720 (commented):
This looks good. The only thing I noticed was that submitting a workflow with a certain visitID updates the visitID in localStorage, but the visitID in localStorage doesn't automatically populate the visitID box on the template page, meaning you still have to type it every time when submitting new workflows.

VictoriaBeilsten-Edmands:
Good point, I've updated it so that ti populated the visit input on the templates page.

PR #810, opened 2025-08-11
Added Frontend formatting using prettier

TBThomas56:
adds functionality to the `yarn lint / precommit` command that will check for formatting issues with no config file being checked.
added functionality to `yarn lint`such that CI stays the same
added `yarn format` to write the format to edit files for formatting


TBThomas56:
> What beautiful formatting! Shall we format .md and .json too?

markdown is already being formatted with the github pages CI, I'll add .json now as well

PR #812, opened 2025-08-12
feat(frontend): filter workflows by template

jacob720:
Adds a filter on the wofklows list page to filter workflows by template.
https://jira.diamond.ac.uk/browse/AP-755

### DIFF
--- a/frontend/workflows-lib/lib/components/workflow/WorkflowListFilterDrawer.tsx
+++ b/frontend/workflows-lib/lib/components/workflow/WorkflowListFilterDrawer.tsx
@@ -50,9 +50,11 @@ export function WorkflowListFilterDisplay({
 }) {
   const statusString = WorkflowStatusToString(filter.workflowStatusFilter);
   const creator = filter.creator;
+  const template = filter.template;
   return (
     <Box sx={{ mb: 2 }}>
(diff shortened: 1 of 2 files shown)


frontend/workflows-lib/lib/components/workflow/WorkflowListFilterDrawer.tsx
```diff
@@ -198,6 +215,30 @@ function WorkflowListFilterDrawer({
...
+                  >
+                    <ClearIcon sx={{ fontSize: 20 }} />
+                  </IconButton>
+                ) : undefined,
+              },
+            }}
+          />
           <Autocomplete
             multiple
             options={[
```
davehadley, line 244: Does WorkflowListFilterDrawer has any associated tests that need to be modified or updated given this new functionality?

davehadley (approved):

jacob720:
> Does this work only with exact templateName matching? If so we probably want a filtered dropdown list with possible values otherwise it could be confusing. We would need template names with which to populate that list.

Yeah it needs to be exact, I agree it would be much nicer to have fuzzy search or a filtered drop down. The question is how can we get a list of template names? I think we briefly talked about a web hook to collect them last week.

PR #814, opened 2025-08-12
fix(frontend): move graphql consts out of component files

VictoriaBeilsten-Edmands:
Moves queries out of component files to allow fast refresh to work correctly.


jacob720 (approved):
Looks good!

PR #817, opened 2025-08-14
feat(graph-proxy): add workflow creator to graph

davehadley:


backend/graph-proxy/src/graphql/workflows.rs
```diff
@@ -715,6 +720,34 @@ impl WorkflowsQuery {
     }
 }
 
+/// Information about the creator of a workflow.
+#[derive(Debug, Clone, SimpleObject, Eq, PartialEq)]
+struct WorkflowCreator {
+    /// An identifier unique to the creator of the workflow.
+    /// Typically this is the creator's Fed-ID.
+    creator_id: String,
```
MaxHerbs, line 728: We should probably decide on what the future is for this. If we're returning a struct with several identifiers incl fedid, prefered-name etc, I might expect creator_id field to be called fedid. There are probably some caveats here regarding service accounts etc. If we're only returning Fedid it's probably worth changing the language

davehadley: The intention is that a future pull request will add additional fields (eg user real name).

My understanding is that in the future it is possible for the creator to not have a fedId. For example some kinds of automated workflows or potentially some users of the post visit analysis service. Hence this choice of name.

VictoriaBeilsten-Edmands (approved):

PR #820, opened 2025-08-14
chore(resources): bump postgres + kyverno resources

MaxHerbs:
Not throttling but ~90% so a bit close for comfort


davehadley (approved):
Do we have any idea why the resource usage is increasing?

PR #830, opened 2025-08-18
refactor(graph-proxy): update science-group filter semantics

MaxHerbs:


backend/graph-proxy/src/graphql/filters.rs
```diff
@@ -0,0 +1,128 @@
...
+
+        self.science_group.generate_labels(&mut label_selectors);
+
+        label_selectors.join(",")
+    }
+}
+
+/// Supported DLS science groups
+#[derive(InputObject, Debug, Default, Clone)]
+pub struct ScienceGroup {
```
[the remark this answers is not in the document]
davehadley: I would have made the available ScienceGroup's an enum.

Then I would have called this struct ScienceGroupFilter that contains a HashMap<ScienceGroup, bool> or something similar.

Then future changes to the ScienceGroup is a one line change to the enum.

backend/graph-proxy/src/graphql/filters.rs
```diff
@@ -0,0 +1,128 @@
...
+    /// Examples Group
+    examples: Option<bool>,
+}
+
+impl GraphFilter for ScienceGroup {
+    fn generate_labels(&self, labels: &mut Vec<String>) {
+        let label_prefix = "workflows.diamond.ac.uk/science-group-";
+
+        let fields = [
+            ("mx", self.mx),
```
davehadley, line 66: if you choose to go down the enum route, these group name strings could be part of the enum

backend/graph-proxy/src/graphql/filters.rs
```diff
@@ -0,0 +1,128 @@
+use async_graphql::InputObject;
+use url::Url;
+
+/// Build labels to apply query to workflows API
+pub trait GraphFilter {
```
davehadley, line 5: I think this trait can be private for now until it's needed in other modules.

backend/graph-proxy/src/graphql/filters.rs
```diff
@@ -0,0 +1,128 @@
...
+    /// The science group owning the template eg imaging
+    science_group: Option<ScienceGroup>,
+}
+
+impl WorkflowTemplatesFilter {
+    /// Generates and applies all the filters
+    pub fn generate_filters(&self, url: &mut Url) {
+        let labels = &self.create_label_selection();
+        url.query_pairs_mut()
+            .append_pair("listOptions.labelSelector", labels);
```
davehadley, line 35: does this behave correctly if labels is an empty string (i.e. no filters were added)?

backend/graph-proxy/src/graphql/filters.rs
```diff
@@ -0,0 +1,128 @@
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::graphql::filters::{ScienceGroup, WorkflowTemplatesFilter};
+
+    #[tokio::test]
+    async fn science_group_filter() {
+        let science_group = ScienceGroup {
```
davehadley, line 86: If possible, consider splitting into three separate tests. This makes debugging easier as you get the result from all 3 asserts at once.

backend/graph-proxy/src/graphql/filters.rs
```diff
@@ -0,0 +1,128 @@
...
+            magnetic_materials: None,
+        };
+        let filters = WorkflowTemplatesFilter {
+            science_group: Some(science_group),
+        };
+        let label_selectors = filters.create_label_selection();
+        assert_eq!(
+            label_selectors,
+            "workflows.diamond.ac.uk/science-group-examples=true"
+        );
```
davehadley, line 126: Consider adding a test for the ScienceGroup{None, None, None, None} case.

davehadley (changes requested):
Thanks for putting this together.

Moving the filter logic into its own module makes sense and makes the codebase easier to maintain and extend.

One thing to keep in mind for future pull requests: this one mixes a feature that changes the public interface (the accepted format for specifying science-group in templates) with some refactoring. It helps the review if these can be two separate commits (so I can just focus on the important changes which is the feature change).

For release-please I think this commit message should be changed to feat.

backend/graph-proxy/src/graphql/filters.rs
MaxHerbs: if no filters are added then the generate_filters function should never be called to start with so that should be fine

PR #838, opened 2025-08-20
Optimise dealing with workflow subscription data

jacob720:
https://jira.diamond.ac.uk/browse/AP-883

We are receiving lots of workflow subscription data, even when workflows are the same. They looked different to the frontend because the tasks are in a random order, so these are sorted before comparing the workflow with the last one received, so we only re-render when the new workflow data is meaningfully different.

### DIFF
--- a/frontend/relay-workflows-lib/lib/utils.ts
+++ b/frontend/relay-workflows-lib/lib/utils.ts
@@ -1,6 +1,11 @@
 import { useState } from "react";
+
 import type { workflowRelayQuery$data } from "./graphql/__generated__/workflowRelayQuery.graphql";
+import { workflowRelaySubscription$data } from "./graphql/__generated__/workflowRelaySubscription.graphql";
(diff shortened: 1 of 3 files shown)


frontend/relay-workflows-lib/lib/utils.ts
```diff
@@ -32,6 +36,74 @@ export function useClientSidePagination<T>(
   };
 }
 
+export function hasTasks(
+  status: WorkflowStatus | null | undefined,
+): status is Extract<WorkflowStatus, { tasks: readonly unknown[] }> {
+  return !!status && "tasks" in status && Array.isArray(status.tasks);
+}
+
+export function workflowsAreEqual(
```
VictoriaBeilsten-Edmands, line 45: Can we replace this with hash comparison?

jacob720: Yeah this worked, I've changed it to use hash comparison

PR #850, opened 2025-08-22
feat(frontend): add mocking

VictoriaBeilsten-Edmands:
Adds a mock mode which can be run with `yarn mock` from `workflows/frontend/dashboard`.
Mocks out queries and keycloak and mocks subscriptions as queries.
Refactors SingleWorkflowView and WorkflowRelay into live, mock, base and top level components. These components prevent the calling of hooks conditionally and should help if we want to conditionally set up subscriptions only for workflows still in progress.
The mocking can be the basis for writing tests in future for components in relay-workflows-lib.
The data in the mocks can be improved when we have better test workflows.

### DIFF
--- a/frontend/dashboard/src/main.tsx
+++ b/frontend/dashboard/src/main.tsx
@@ -47,16 +47,25 @@ const router = createBrowserRouter([
   },
 ]);
 
+async function startMockingIfConfigured() {
+  if (import.meta.env.VITE_ENABLE_MOCKING === "true") {
+    const { worker } = await import("./mocks/browser");
+    return worker.start();
+  }
+}
+
 const root = createRoot(document.getElementById("root") as Element);
 
-void getRelayEnvironment().then((environment) => {
-  root.render(
-    <RelayEnvironmentProvider environment={environment}>
-      <StrictMode>
-        <ThemeProvider theme={DiamondTheme} defaultMode="light">
-          <RouterProvider router={router} />
-        </ThemeProvider>
-      </StrictMode>
-    </RelayEnvironmentProvider>,
-  );
+void startMockingIfConfigured().then(() => {
+  void getRelayEnvironment().then((environment) => {
+    root.render(
+      <RelayEnvironmentProvider environment={environment}>
+        <StrictMode>
+          <ThemeProvider theme={DiamondTheme} defaultMode="light">
+            <RouterProvider router={router} />
+          </ThemeProvider>
+        </StrictMode>
+      </RelayEnvironmentProvider>,
+    );
+  });
 });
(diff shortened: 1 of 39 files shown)


frontend/relay-workflows-lib/lib/components/SingleWorkflowView.tsx
```diff
@@ -1,170 +1,18 @@
...
-      </Box>
-      {tasknames && (
-        <TaskInfo
-          artifactList={artifactList}
-          onArtifactHover={onArtifactHover}
-        />
-      )}
-      {workflowData && <WorkflowInfo workflow={workflowData.workflow} />}
-    </>
+export default function SingleWorkflowView(props: SingleWorkflowViewProps) {
```
davehadley, line 12: does typescript have a mechanism to swap in the mock code at test time so you don't have to modify the production component like this? that would be nicer if it is easy to achieve.

frontend/dashboard/src/main.tsx
```diff
@@ -47,16 +47,25 @@ const router = createBrowserRouter([
   },
 ]);
 
+async function enableMocking() {
```
davehadley, line 50: Suggest function name chance. Name implies that it always enables mocks. Reader might think this is a bug if they see it in a production context without reading the content of the function. Suggest: startMockingIfConfigured

frontend/dashboard/src/mocks/mockKeycloak.ts
```diff
@@ -0,0 +1,24 @@
...
+  const header = btoa(JSON.stringify({ alg: "HS256", typ: "JWT" }));
+  const body = btoa(JSON.stringify(payload));
+  return `${header}.${body}.signature`;
+};
+
+const mockToken = createMockJwt({
+  sub: "mock-user",
+  preferred_username: "mockuser",
+  email: "mock@diamond.ac.uk",
+  exp: Math.floor(Date.now() / 1000) + 3600,
```
davehadley, line 11: To make tests deterministic, maybe set to Number.MIN_SAFE_INTEGER and Number.MAX_SAFE_INTEGER

frontend/relay-workflows-lib/lib/components/LiveSingleWorkflowView.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+  const subscriptionData: GraphQLSubscriptionConfig<WorkflowRelaySubscriptionType> =
+    useMemo(
+      () => ({
+        subscription: workflowRelaySubscription,
+        variables: { visit, name: workflowName },
+        onNext: (res) => {
+          setWorkflowData(res ?? null);
+        },
+        onError: () => {},
```
davehadley, line 28: should we log something to the console here?

frontend/dashboard/src/main.tsx
davehadley: Is it a problem in typescript that one branch returns a promise and the else branch returns undefined?

davehadley (dismissed):

frontend/dashboard/src/main.tsx
VictoriaBeilsten-Edmands: The function isn't awaited and the result is thrown away with the `void` so it's fine here.

`void enableMocking().then(() => {`

PR #856, opened 2025-08-27
feat(examples): provide test template for developement

jacob720:
https://jira.diamond.ac.uk/browse/AP-878
Test template to help with frontend development.
 Will extend in another PR to add more artifacts such as a hdf5 file


examples/conventional-templates/test-template.yaml
```diff
@@ -0,0 +1,294 @@
+apiVersion: argoproj.io/v1alpha1
+
+kind: Workflow
+
+metadata:
+  name: test-template
```
MaxHerbs, line 6: Probably worth having a more descriptive name

PR #858, opened 2025-08-27
Lint workflows by labels and annotations

MaxHerbs:
Depends on #857 

This is supporting 'critical' (ie failing) labels. As things grow this should scale to add 'recommended' and 'depreciated' as well. I'm open to input on the terminology/naming for the config - and whether it should be decided at compile time or should download the latest when linting. It made sense at the time of writing, but best we decide now rather than changing things down the line.

### DIFF
--- a/workflows-cli/src/linter/linter_labels.rs
+++ b/workflows-cli/src/linter/linter_labels.rs
@@ -1,0 +1,338 @@
+use crate::linter::base_linting::{Linter, get_manifest};
+use serde::Deserialize;
+use serde_yaml::Value;
+use std::{
+    fmt::{self},
+    path::Path,
+    vec,
+};
+
+const RULES_YAML: &str = include_str!("../../../.workflow_metadata_ruleset.yaml");
+
+pub struct LabelChecker {
+    rules: Rules,
+}
+
+impl Linter for LabelChecker {
+    fn lint(target: &Path) -> Result<Vec<String>, String> {
+        let manifest = get_manifest(target)?;
+        let label_checker = LabelChecker::new()?;
+
+        let metadata = manifest
+            .as_mapping()
+            .ok_or("Invalid manifest: expected a YAML object (key-value pairs), but got a different type.")?
+            .get("metadata")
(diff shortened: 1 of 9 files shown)


workflows-cli/src/linter/linter_labels.rs
```diff
@@ -0,0 +1,330 @@
...
+        match self {
+            CriticalRule::Match(m) => m.check(labels, annotations),
+            CriticalRule::Stem(s) => s.check(labels, annotations),
+        }
+    }
+}
+
+impl RuleChecker for Match {
+    fn check(&self, labels: &Value, annotations: &Value) -> Result<(), String> {
+        let exists = match self.location {
```
davehadley, line 118: This enforces that the key is present in the expected location.

Should a match rule enforce that the key is *not present* in the wrong location?

.workflows-labels.yaml
```diff
@@ -0,0 +1,11 @@
+critical:
```
davehadley, line 1: suggest a more descriptive name for this file

Maybe: `default_rule_set_for_workflow_labels_and_annotations.yaml`?

workflows-cli/src/linter/linter_labels.rs
```diff
@@ -0,0 +1,330 @@
...
+    }
+}
+
+#[derive(Debug, Deserialize)]
+#[serde(tag = "type")]
+enum CriticalRule {
+    #[serde(rename = "stem")]
+    Stem(Stem),
+    #[serde(rename = "match")]
+    Match(Match),
```
davehadley, line 65: `Match` might be more descriptively labelled `requiredKey`?

workflows-cli/src/linter/linter_labels.rs
```diff
@@ -0,0 +1,330 @@
...
+                Location::Label => labels.get(&formatted_option),
+            };
+
+            if let Some(val) = value {
+                entries.push((formatted_option.clone(), val));
+            }
+        }
+
+        if entries.is_empty() {
+            let msg = format!("Expected {}-<GROUP> in {}", self.stem, self.location);
```
davehadley, line 148: Stem rule code is generic over more than just `science-group` but error message imples that the `leaf` of the key is a `<GROUP>`.

davehadley (approved):

PR #864, opened 2025-08-29
Fix bugs caused by same task names

jacob720:
Fixes bugs caused by tasks having the same name.
A useful workflow for testing this: http://localhost:5173/workflows/mg36964-1/example-template-zjjjv

I also made the task fill colour more vibrant as it was quite hard to see.


VictoriaBeilsten-Edmands (approved):
Looks good.

PR #873, opened 2025-09-03
feat(examples): added template for benchmarking subscriptions

NatLeung96:
Added a template for testing subscription benchmarks.

### DIFF
--- a/examples/conventional-templates/subsrciption-benchmark.yaml
+++ b/examples/conventional-templates/subsrciption-benchmark.yaml
@@ -1,0 +1,91 @@
+apiVersion: argoproj.io/v1alpha1
+kind: ClusterWorkflowTemplate
+metadata:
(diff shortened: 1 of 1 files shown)


examples/conventional-templates/subsrciption-benchmark.yaml
```diff
@@ -0,0 +1,91 @@
...
+      command: ["python"]
+      source: |
+        print("This is operation-ten")
+  - name: workflow-entry
+    dag:
+      tasks:
+      - name: operation-01
+        template: operation-one
+      - name: operation-02
+        template: operation-two
```
VictoriaBeilsten-Edmands, line 74: Each task needs to define which task it depends on to run one after the other.
```suggestion
        template: operation-two
        dependencies: [operation-01]
```

PR #877, opened 2025-09-04
feat(frontend): changed navigation link to workflow name page

NatLeung96:
I believe the issue in AP-886 was due to useResolvedPath. I have instead defined a full path to the workflow page in the icon link.


VictoriaBeilsten-Edmands:
Looks like a change that got introduced in a previous PR.

PR #883, opened 2025-09-09
Nl/show workflow status after submission

NatLeung96:
This PR adds a live workflow status icon to the submission message. I'm not 100% is this is the best implementation. I would therefore greatly appreciate any feedback on this one.


VictoriaBeilsten-Edmands (changes requested):
I would move the subscription out of TemplateView into SubmittedMessageList. The subscription needs to be able to track multiple different workflows. At the moment, the status of the last submitted workflow is used for all workflows in the list. 
For example, if you have workflows that have finished in the list and then submit another one, the new status is shown for all.
Screenshot 2025-09-09 at 15 23 22

After submitting another workflow
Screenshot 2025-09-09 at 15 23 30

frontend/relay-workflows-lib/lib/components/SubmittedMessagesList.tsx
```diff
@@ -76,16 +93,53 @@ const renderSubmittedMessage = (
   }
 };
 
-interface SubmittedMessagesListProps {
-  submissionResults: (
+interface SubscribeAndRenderPropsList {
+  r:
```
VictoriaBeilsten-Edmands, line 97: Maybe change these props from `r` and `i` to `result` and `index` to make it a bit easier to follow.

frontend/relay-workflows-lib/lib/components/SubmittedMessagesList.tsx
```diff
@@ -35,6 +47,11 @@ const renderSubmittedMessage = (
             <Typography>
               Successfully submitted{" "}
               <Link to={`/workflows/${r.message}`}>{r.message}</Link>
+              {workflowData
```
VictoriaBeilsten-Edmands, line 50: Can this icon be vertically centred with the text?

frontend/relay-workflows-lib/lib/components/SubmittedMessagesList.tsx
```diff
@@ -103,14 +157,25 @@ const SubmittedMessagesList: React.FC<SubmittedMessagesListProps> = ({
...
-      {submissionResults.length > 0 && (
+      {submittedData.length > 0 && (
         <>
           <Paper elevation={1} sx={{ p: 2 }}>
             <Typography variant="h6" sx={{ textAlign: "center" }}>
               Submissions
             </Typography>
           </Paper>
-          {submissionResults.map((r, i) => renderSubmittedMessage(r, i))}
+          {submittedData.map((data, i) =>
```
VictoriaBeilsten-Edmands, line 167: Can this be a set width so that all of the status icons are vertically aligned with one another, including when one of the submissions has an error?

VictoriaBeilsten-Edmands (changes requested):

PR #887, opened 2025-09-10
feat(frontend): add search filter to template list

JamesDoingStuff:
Adds a search bar to the template page that queries the name, title and description of the templates and updates the list accordingly.

### DIFF
--- a/frontend/relay-workflows-lib/lib/components/TemplatesList.tsx
+++ b/frontend/relay-workflows-lib/lib/components/TemplatesList.tsx
@@ -1,10 +1,11 @@
-import { ChangeEvent } from "react";
+import { ChangeEvent, useMemo, useState } from "react";
 import { graphql } from "relay-runtime";
 import { useLazyLoadQuery } from "react-relay/hooks";
 import { Box, Pagination } from "@mui/material";
-import { TemplateCard } from "workflows-lib";
+import { TemplateCard } from "workflows-lib/lib/components/template/TemplateCard";
 import { TemplatesListQuery } from "./__generated__/TemplatesListQuery.graphql";
 import { useClientSidePagination } from "../utils";
+import TemplateSearchField from "workflows-lib/lib/components/template/TemplateSearchField";
 
 const templatesListQuery = graphql`
   query TemplatesListQuery {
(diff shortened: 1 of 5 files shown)


VictoriaBeilsten-Edmands:
I would do away with the submit button, but add an 'X' icon to clear the input of the search.

frontend/relay-workflows-lib/lib/components/TemplatesList.tsx
```diff
@@ -22,41 +23,78 @@ const templatesListQuery = graphql`
...
+  useEffect(() => {
+    const templates = data.workflowTemplates.nodes;
+    const filteredTemplates = [];
+    if (search === "") {
+      setFilteredTemplates(templates);
+    }
+
+    const upperSearch = search.toUpperCase();
+
+    for (let i = 0; i < templates.length; i++) {
```
VictoriaBeilsten-Edmands, line 55: This can be simplified with
` templates.filter`

frontend/relay-workflows-lib/lib/components/TemplatesList.tsx
```diff
@@ -22,41 +23,78 @@ const templatesListQuery = graphql`
 
 export default function TemplatesList() {
   const data = useLazyLoadQuery<TemplatesListQuery>(templatesListQuery, {});
-  const templates = data.workflowTemplates.nodes;
+  const [search, setSearch] = useState("");
+  const [filteredTemplates, setFilteredTemplates] = useState(
```
VictoriaBeilsten-Edmands, line 27: This doesn't need to be a state because it's essentially derived from another state `search`

frontend/relay-workflows-lib/lib/components/TemplatesList.tsx
```diff
@@ -22,41 +23,78 @@ const templatesListQuery = graphql`
...
     totalPages,
     paginatedItems: paginatedPosts,
-  } = useClientSidePagination(templates, 10);
+  } = useClientSidePagination(filteredTemplates, 10);
 
   const handlePageChange = (_event: ChangeEvent<unknown>, page: number) => {
     setPageNumber(page);
   };
 
+  useEffect(() => {
```
VictoriaBeilsten-Edmands, line 46: then instead of a `useEffect` here, you can use a `useMemo` e.g.
```
const filteredTemplates = useMemo(() => {
  const upperSearch = search.toUpperCase();
  return templates.filter(template => 
  ...
  }, [search, templates])
```

frontend/relay-workflows-lib/lib/components/TemplatesList.tsx
```diff
@@ -22,41 +23,78 @@ const templatesListQuery = graphql`
...
+        width="100%"
       >
-        <Pagination
-          count={totalPages}
-          page={pageNumber}
-          onChange={handlePageChange}
-          showFirstButton
-        />
+        {paginatedPosts.map((node, index) => (
+          <TemplateCard key={index} template={node} />
```
VictoriaBeilsten-Edmands, line 78: template.name should be unique here in place of index

VictoriaBeilsten-Edmands (changes requested):

PR #903, opened 2025-09-15
feat(frontend): add workflow error message to ui

NatLeung96:


MaxHerbs:
This is more style than anything - can we have the box not appear at all if there is no error? I would only expect an error in the message field in ~1% of workflows so it's not a common pattern but needs highlighting when it does occur

PR #946, opened 2025-10-06
Vbe/relay fragments

VictoriaBeilsten-Edmands:


frontend/relay-workflows-lib/lib/components/WorkflowInfo.tsx
TBThomas56, line 1: More of a question for understanding. What is the fragment$key and fragment$data used for? and is that different now that we use the useFragment function?

PR #985, opened 2025-10-22
feat(charts): enable tolerations for workflows nodes by default

davehadley:


davehadley:
Thanks for catching that @JamesDoingStuff. The issue was that the cs05r_gpfs:PreferNoSchedule weight seems to outweigh the workflows nodeAffinity weight, even when set to 100.

I have updated it to tolerate cs05r_gpfs taint. That seems to fix the issue.

JamesDoingStuff (approved):
You're welcome, checked again and all running on the right nodes now :+1:

PR #1000, opened 2025-10-28
chore(charts): increase kyverno CPU and memory limits in staging

davehadley:


TBThomas56 (commented):
would this bump be enough? Should the requests limits be increased for both of these as well?

TBThomas56 (approved):

PR #1028, opened 2025-11-04
fix(charts): hot-fix high volume of kyverno logs causing graylog disruption

davehadley:


davehadley:
See: https://diamondlightsource.slack.com/archives/C08NYJSGMFD/p1762165700348629

TBThomas56 (approved):

PR #1042, opened 2025-11-13
chore(charts): enable ingress for grafana.staging.workflows.diamond.ac.uk

davehadley:


TBThomas56 (approved):
LGTM
Only thing I wanted to ask is if we want to sync secrets at the cluster level? Or is this something we want to try in the future?

TBThomas56:
> > LGTM Only thing I wanted to ask is if we want to sync secrets at the cluster level? Or is this something we want to try in the future?
> 
> I'm not sure what you mean.
> 
> If you mean, have the same secret for argus and pollux, that is not possible. Keycloak admins told me that staging and prod must have different clients and that means different client IDs and client secrets.
> 
> Long term I would like to eliminate the monitoring stack and migrate the the centrally provided one by cloud team.

When I mean sync the secrets. We have moved a lot of the secrets in the cluster e.g. "argo-server-sso" in the workflows-cluster chart. This then syncs the secrets into the cluster. Was asking if a similar thing should be done with the secrets here.

PR #1089, opened 2026-01-09
fix(charts): prevent race-condition in kyverno chainsaw tests causing spurious failures

davehadley:


davehadley:
For context see: https://diamondlightsource.slack.com/archives/C06N43M7JP3/p1763143391265749?thread_ts=1763137059.279019&cid=C06N43M7JP3

JamesDoingStuff (approved):

PR #1229, opened 2026-04-09
Fix re-submission of workflows with array parameters

JamesDoingStuff:
[AP-1072](https://jira.diamond.ac.uk/browse/AP-1072)
Tweaks graph-proxy to treat workflow parameters as `Value`s instead of strings, allowing arrays to be returned by queries

### DIFF
--- a/backend/graph-proxy/src/graphql/workflows.rs
+++ b/backend/graph-proxy/src/graphql/workflows.rs
@@ -11,6 +11,7 @@ use async_graphql::{
 use aws_sdk_s3::presigning::PresigningConfig;
 use axum_extra::headers::{authorization::Bearer, Authorization};
 use chrono::{DateTime, Utc};
+use serde_json::{from_str, Value};
 use std::{collections::HashMap, ops::Deref, path::Path};
 use tracing::{debug, instrument};
 use url::Url;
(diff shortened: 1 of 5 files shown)


backend/graph-proxy/src/graphql/workflows.rs
```diff
@@ -92,18 +93,18 @@ impl Workflow {
...
-        let arguments = self.manifest.spec.arguments.as_ref();
-
-        if let Some(args) = arguments {
-            let params = &args.parameters;
-            let mut param_map: HashMap<&str, Option<&str>> = HashMap::new();
-            params.iter().for_each(|this_parameter| {
-                param_map.insert(&this_parameter.name, this_parameter.value.as_deref());
-            });
-            return Some(param_map);
+    async fn parameters(&self) -> Option<HashMap<&str, Value>> {
```
davehadley, line 96: I think this will cause a breaking change to the GraphQL API.

We are the only consumers of this API right now so it is fine, but we need to be careful with these kinds of changes in the future, once others start depending upon it.

davehadley (approved):

backend/graph-proxy/src/graphql/workflows.rs
JamesDoingStuff: Noted, good point :+1:

PR #1231, opened 2026-04-10
feat: graph-proxy requires authn for all querys/mutations

davehadley:


backend/graph-proxy/src/validate_token.rs
```diff
@@ -0,0 +1,319 @@
...
+
+    use super::{TokenValidator, ValidatedAuthToken, ValidationMethod};
+
+    #[tokio::test]
+    #[rstest]
+    #[case(ValidationMethod::Introspection)]
+    #[case(ValidationMethod::Jwt)]
+    async fn test_token_validator(#[case] method: ValidationMethod) -> anyhow::Result<()> {
+        if std::env::var("WORKFLOWS_DEV_CONTAINER").is_ok() {
+            eprintln!("Skipping test: test containers don't work inside VSCode dev container");
```
TBThomas56, line 273: I think you can use testcontainers in a Dev Container. From my understanding there is one in Auth-Daemon and there is a GitHub issue here: https://github.com/testcontainers/testcontainers-java/discussions/9793

backend/graph-proxy/src/validate_token.rs
```diff
@@ -0,0 +1,319 @@
...
+        match self {
+            ValidatedAuthToken::Valid(authorization) => Some(authorization),
+            ValidatedAuthToken::Invalid => None,
+            ValidatedAuthToken::Missing => None,
+            ValidatedAuthToken::Failed(_) => None,
+        }
+    }
+}
+
+/// Checks OIDC access_tokens are valid
```
TBThomas56, line 59: Just a question. Will this be using any auth features we had in the oidc-bff example. Just to understand if next steps would be to connect it to our auth-backend system?

backend/graph-proxy/src/graphql/auth_guard.rs
```diff
@@ -0,0 +1,48 @@
...
+        }
+    }
+}
+
+impl Display for AuthErrorCode {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let value = match self {
+            AuthErrorCode::Unauthenticated => "UNAUTHENTICATED",
+        };
+        write!(f, "{}", value)
```
TBThomas56, line 46: would it be better to have write! or tracing?

backend/graph-proxy/src/validate_token.rs
```diff
@@ -0,0 +1,319 @@
...
+};
+use serde_json::Value;
+
+#[derive(Debug, Clone, PartialEq)]
+/// Method of token validation
+pub enum ValidationMethod {
+    /// Use OIDC provider introspection end-point
+    #[allow(dead_code)]
+    Introspection,
+    /// Use local validation of the JWT
```
TBThomas56, line 27: Just for understanding. Why are we using Jwt here and later trying to get a token from Identity?

backend/graph-proxy/src/graphql/mod.rs
```diff
@@ -58,7 +61,7 @@ pub struct Query(NodeQuery, WorkflowsQuery, WorkflowTemplatesQuery);
 #[derive(Debug, Clone, Default)]
 pub struct NodeQuery;
 
-#[Object]
+#[Object(guard = "AuthGuard")]
```
TBThomas56, line 64: for understanding: what is appending when we are changing #[Object] to #[Object(guard = "AuthGard")]. Not following here

backend/graph-proxy/src/graphql/workflow_templates.rs
```diff
@@ -147,7 +147,7 @@ impl WorkflowTemplate {
 #[derive(Debug, Clone, Default)]
 pub struct WorkflowTemplatesQuery;
 
-#[Object]
+#[Object(guard = "AuthGuard")]
 impl WorkflowTemplatesQuery {
```
TBThomas56, line 151: Just checking, this has been checked to ensure that all Workflows and Templates are visible. I do not know if it is me, but I was not able to see either in staging

backend/graph-proxy/src/validate_token.rs
davehadley: It would require docker-in-docker (which may not work on podman on Diamond desktops).
So I have disabled it for now. The test runs outside the dev container and in github CI so it will get run before a merge.

backend/graph-proxy/src/validate_token.rs
davehadley: If TokenValidator is needed elsewhere that can be part of a later re-factor.

backend/graph-proxy/src/graphql/auth_guard.rs
davehadley: write! is mutating `f` to store the string `value`.

That then later becomes part of the response to the graphQL query.

backend/graph-proxy/src/validate_token.rs
davehadley: There are two ways to check a Keycloak token:

(1) Send the token to the Keycloak introspection endpoint
(2) Check the JWT matches its signature and the signature matches the public key of the Keycloak instance

(2) entirely local so it is faster and puts no loads on the Keycloak server.

(1) is able to check if the keycloak token has been revoked. Recommendation seems to be to use this method for security sensitive operations eg

For now I have configured graph-proxy to only do (2) but in future we may need (1) for some fields or mutations.

backend/graph-proxy/src/graphql/mod.rs
davehadley: This is a field guard:
https://async-graphql.github.io/async-graphql/en/field_guard.html

This means: "if someone tries to access this field, run the `check` method of `AuthGuard`.
If `check` returns `Ok(())` then let the person access that field.
If `check` returne `Err` then don't let them access that field.

Applying it to the `#[Object]` is the same as applying it to every field.

backend/graph-proxy/src/graphql/workflow_templates.rs
davehadley: It failed on staging because the audiences are different. I am fixing this now and will retest on staging before merge.

davehadley:
Subscriptions currently don't work. Reducing the scope for this PR to just queries and mutations.

TBThomas56 (approved):

PR #1232, opened 2026-04-10
K6 chart and test merge request

MahirAbbas:
Merge k6 chart as well as test into main


TBThomas56:
@MahirAbbas  Would separate the commits to the Deployment. And anything in the Smoke Test. 
CI is breaking because you need to change your commit messages to follow conventional commits (https://www.conventionalcommits.org/en/v1.0.0/#summary) and the App chart Version needs updating

PR #1237, opened 2026-04-13
feat(charts): k6 deployment in pollux

MahirAbbas:


charts/apps/templates/k6-operator.yaml
```diff
@@ -0,0 +1,35 @@
...
+apiVersion: argoproj.io/v1alpha1
+kind: Application
+metadata:
+  name: k6-operator
+  namespace: argocd
+  annotations:
+    argocd.argoproj.io/sync-wave: "1"
+spec:
+  destination:
+    namespace: {{ .Values.k6Operator.namespace }}
```
TBThomas56: I would add the target namespace in your values.yaml file so that k6 is deployed in its own namespace instead of the argocd one

TBThomas56 (changes requested):
All good! Just add the namespace.

charts/apps/templates/k6-operator.yaml
MahirAbbas: yep just added in values.yaml. lmk if you're unhappy with how I did it. If not, if everything looks good then hopefully we can merge.

TBThomas56 (approved):

PR #1238, opened 2026-04-13
feat(charts): add k6 smoke test

MahirAbbas:


charts/k6-operator/tests/ping-graph.ts
```diff
@@ -0,0 +1,97 @@
+import http, { RefinedResponse, ResponseType } from 'k6/http';
+import { Options } from 'k6/options';
+
+const diamondToken = '';
```
TBThomas56: Question: is this the next actionable step? You want to get the tokens in? Maybe have it set to get this from an environmental variable if that is how you are going to access this?

TBThomas56 (changes requested):

TBThomas56 (approved):

PR #1243, opened 2026-04-15
Tbt/auth refactor

TBThomas56:


backend/auth-core/src/middleware/inject_token.rs
```diff
@@ -0,0 +1,72 @@
...
+            ))
+        }
+    } else {
+        Ok(next.run(req).await)
+    }
+}
+
+/// Axum middleware that extracts `State<Arc<S>>` and delegates to [`inject_token_with`].
+/// Use this when the store is shared application state (e.g. auth-daemon's RouterState).
+pub async fn inject_token<S>(
```
davehadley, line 63: Commenting as I go along, so haven't go to where these are used:

do both of these need to be public?

backend/auth-core/src/middleware/traits.rs
```diff
@@ -0,0 +1,25 @@
+use async_trait::async_trait;
+use axum::http::StatusCode;
+
+use crate::Result;
+
+// How tokens ore loaded, saved and refreshed
```
davehadley, line 6: typo ore

backend/auth-core/src/database.rs
```diff
@@ -0,0 +1,139 @@
...
+    // Check when token expires
+    let expires_at_utc = match row.expires_at {
+        Some(dt) => to_utc(dt),
+        None => {
+            // If not stored, decide on a policy: here we treat as "no expiry"
+            // You can choose to error instead.
+            Utc::now()
+        }
+    };
+    if expires_at_utc < Utc::now() {
```
davehadley, line 67: What is the intentional behavior here when expire_at is None?

Suppose the following:

`expires_at_utc` is set to `Utc::now()`.

The OS puts the process to sleep for a few seconds.

It wakes up and continues and performs the `expires_at_utc < Utc::now()` check?

Does this do the correct thing?

backend/oidc-bff/src/login.rs
```diff
@@ -0,0 +1,46 @@
...
+    // Generate the full authorization URL.
+    let (auth_url, csrf_token, nonce) = oidc_client
+        .authorize_url(
+            CoreAuthenticationFlow::AuthorizationCode,
+            CsrfToken::new_random,
+            Nonce::new_random,
+        )
+        // Set the desired scopes.
+        .add_scope(Scope::new("openid".to_string()))
+        .add_scope(Scope::new("offline_access".to_string()))
```
davehadley, line 34: we need scope `posix-uid` as well now

davehadley: also probably `fedid`

backend/oidc-bff/config.yaml
```diff
@@ -0,0 +1,11 @@
+client_id: "workflows-ui-dev"
+client_secret: ""
+oidc_provider_url: "https://authn.diamond.ac.uk/realms/master"
```
davehadley, line 3: should be identity

charts/apps/staging-values.yaml
```diff
@@ -57,7 +57,7 @@ sealedsecrets:
 
 workflows:
   enabled: true
-  targetRevision: HEAD
+  targetRevision: drh/bff-pkce-dev
```
davehadley, line 60: this needs to be removed

charts/apps/values.yaml
```diff
@@ -122,7 +122,7 @@ sealedsecrets:
 
 workflows:
   enabled: true
-  targetRevision: HEAD
+  targetRevision: drh/bff-pkce-dev
```
davehadley, line 125: this needs to be removed

charts/apps/values.yaml
```diff
@@ -134,7 +134,7 @@ graphProxy:
 
 dashboard:
   enabled: true
-  targetRevision: HEAD
+  targetRevision: drh/bff-pkce-dev
```
davehadley, line 137: this needs to be removed

charts/dashboard/templates/oidc-bff.yaml
```diff
@@ -0,0 +1,70 @@
+apiVersion: v1
```
davehadley, line 1: This would be where I suggest to lock this feature behind a feature flag:
```
{- if .Values.auth.bff.enabled }}
....
{{- end }}
```

.github/workflows/_auth_daemon_container.yaml
```diff
@@ -34,7 +34,7 @@ jobs:
           images: ${{ env.IMAGE_REPOSITORY }}
           tags: |
             type=raw,value=${{ steps.tags.outputs.version }}
-            type=raw,value=latest
+            type=raw,value=staging
```
davehadley, line 37: I don't understand why this is pointing at staging?

charts/apps/values.yaml
```diff
@@ -137,6 +137,7 @@ dashboard:
   targetRevision: HEAD
   extraValuesFiles: []
   valuesObject: {}
+  namespace: dashboard
```
davehadley, line 140: I think that dashboard was merged into the workflows namespace for a reason? so that it could call the database IIRC? is that not neccessary?

charts/dashboard/staging-values.yaml
```diff
@@ -13,7 +13,32 @@ ingress:
...
+          service:
+            name: oidc-bff
+            port: 80
 
   tls: true
   secretName: dashboard-tls-cert
   host: staging.workflows.diamond.ac.uk
+
+oidcBff:
+  enabled: true
```
davehadley, line 35: should it be enabled by default? don't you want it off for now?

charts/dashboard/staging-values.yaml
```diff
@@ -13,7 +13,32 @@ ingress:
...
+  image:
+    registry: ghcr.io
+    repository: diamondlightsource/workflows-oidc-bff
+    tag: "davehadley"
+    digest: "sha256:e259494171d5115d060dd9f3d9da971312d39e9afabd91478c0b7a49c7c5f6d3"
```
davehadley, line 40: why this specific digest?

charts/workflows-cluster/charts/secrets/templates/postgres.yaml
```diff
@@ -19,21 +19,6 @@ spec:
 ---
 apiVersion: bitnami.com/v1alpha1
 kind: SealedSecret
-metadata:
```
davehadley, line 22: does these changes to the passwords work with commit 793eed24

charts/workflows-cluster/staging-values.yaml
```diff
@@ -68,15 +68,15 @@ vcluster:
           byName:
             "/letsencrypt-argo-cd-staging-workflows-diamond-ac-uk": "argocd/argo-cd-tls-cert"
             "/letsencrypt-argo-workflows-staging-workflows-diamond-ac-uk": "workflows/workflows-tls-cert"
-            "/letsencrypt-staging-workflows-diamond-ac-uk": "dashboard/dashboard-tls-cert"
+            "/letsencrypt-staging-workflows-diamond-ac-uk": "workflows/dashboard-tls-cert"
```
davehadley, line 71: there seems to be a namespace discrepancy, see comment about about `dashboard` namespace?

charts/workflows-cluster/staging-values.yaml
```diff
@@ -110,6 +110,7 @@ authenticationConfiguration:
     - issuer:
         url: https://authn.diamond.ac.uk/realms/master
         audiences:
+          - workflows-dashboard-staging
```
davehadley, line 113: What is this audience for?

charts/workflows/templates/_helpers.tpl
```diff
@@ -0,0 +1,28 @@
+{{/*
```
davehadley, line 1: check this is compatible with 793eed24

I think there is some mixing of the old way of dealing with secrets with the new way.

I don't know if a rebase is needed?

charts/workflows/staging-values.yaml
```diff
@@ -19,6 +19,10 @@ oauth2-proxy:
               - aud
             emailClaim: email
             userIDClaim: fedid
+            extraAudiences:
+              - workflows-dashboard-staging
```
davehadley, line 23: is this audience needed?

charts/workflows/values.yaml
```diff
@@ -180,6 +180,10 @@ oauth2-proxy:
               - aud
             emailClaim: email
             userIDClaim: fedid
+            extraAudiences:
+              - workflows-dashboard-staging
```
davehadley, line 184: is this audience needed?

examples/conventional-templates/workflow-auth.yaml
```diff
@@ -0,0 +1,66 @@
+apiVersion: argoproj.io/v1alpha1
```
davehadley, line 1: I suggest leaving this out of this PR. I think there's no way to stop is appearing in the production dashboard?

examples/conventional-templates/workflow-of-workflows.yaml
```diff
@@ -0,0 +1,74 @@
+apiVersion: argoproj.io/v1alpha1
```
davehadley, line 1: again, suggest leaving examples out.

If you want to keep them put them somewhere that doesn't get synced into the production dashboard.

charts/apps/values.yaml
TBThomas56: So this was a way of separating the production from using the current version and staging using the new version. Eventually prod will also have workflows namespace once we point it there

charts/workflows-cluster/staging-values.yaml
TBThomas56: See response there

charts/dashboard/staging-values.yaml
TBThomas56: unsure. was what was there in the proof-of-concept

charts/dashboard/staging-values.yaml
TBThomas56: want it off for prod and on for staging

.github/workflows/_auth_daemon_container.yaml
TBThomas56: auth-daemon very different from what was there is currently. To have things separately, the current tag is set to be at staging. Any further changes should be built to this while we are finalizing things and when we move back to a condition that can be used in prod, we can have the tag changed back to latest.

Prod focuses on latest. Staging looks for staging tag

This was what I could think of in terms of having separate ones. Very happy for suggestions

backend/auth-core/src/middleware/inject_token.rs
TBThomas56: Yes. This has slightly different usecases where one is used by auth-daemon (inject_token) and oidc-bff(inject_token_with)

TBThomas56: added more clear docstrings to show this

charts/workflows/templates/_helpers.tpl
TBThomas56: good catch!

PR #1246, opened 2026-04-17
revert(charts): "chore(charts): upgrade vcluster from 0.28.0 to 0.33.0"

davehadley:
This reverts commit 2a9e06e45ba5809b15aece0664560f130e4ef480.


davehadley:
Reverting this version upgrade as it coincided with the pollux outage. We were only able to recover by deleting the etcd PVCs and rolling back the vCluster version. Further investigation is needed before we can safely upgrade the vCluster version.

JamesDoingStuff (approved):

PR #1252, opened 2026-04-21
client credential flow

MahirAbbas:


charts/k6-operator/templates/k6-testrun.yaml
```diff
@@ -13,11 +13,18 @@ spec:
...
             name: graph-proxy-k6-auth
-            key: GRAPH_PROXY_BEARER_TOKEN
+            key: KEYCLOAK_CLIENT_ID
+      - name: KEYCLOAK_CLIENT_SECRET
+        valueFrom: 
+          secretKeyRef: 
+            name: graph-proxy-k6-auth
+            key: KEYCLOAK_CLIENT_SECRET
+      - name: KEYCLOAK_TOKEN_URL
+        value: https://identity.diamond.ac.uk/realms/dls/protocol/openid-connect/token
```
davehadley: FYI the initial token that you were given in https://jira.diamond.ac.uk/servicedesk/customer/portal/5/UASHD-4191
was for
https://identity-test.diamond.ac.uk/realms/dls

There are three Keycloak servers at Diamond: identity-test, identity-dev, identity (production). You need to use one that matches the client that you were given.

Note the workflows staging cluster only works with the identity servers listed here:

https://github.com/DiamondLightSource/workflows/blob/d80f517617359154bf0e4d4c6a80e4b4198f092a/charts/workflows-cluster/staging-values.yaml#L86

If you want your system to work with identity-test then the staging cluster needs to be modified to accept that server as well.

PR #1259, opened 2026-04-27
fix(sessionspaces): temporarily disable sessionspaces graylog output due to high volume

davehadley:


davehadley:
See: https://diamondlightsource.slack.com/archives/C06N43M7JP3/p1777288918161439

VictoriaBeilsten-Edmands (approved):

PR #1275, opened 2026-05-01
feat(backend): added test sessionspaces for development

TBThomas56:


backend/sessionspaces/src/main.rs
```diff
@@ -66,12 +69,22 @@ async fn main() {
...
+        .unwrap();
+
     let mut current_sessions = Sessions::default();
     let mut update_interval = interval(args.update_interval.into());
     loop {
         update_interval.tick().await;
         match Sessions::fetch(&ispyb_pool, &mut ldap_connection).await {
             Ok(mut new_sessions) => {
+                if let Some(ref statics) = static_sessions {
+                    statics.merge_into(&mut new_sessions);
```
davehadley: what prevents these "static" sessions being repeatedly re-applied each loop?

backend/sessionspaces/src/main.rs
```diff
@@ -66,12 +69,22 @@ async fn main() {
         }
         builder.build()
     };
+    let static_sessions = args
+        .static_sessions
+        .as_deref()
+        .map(StaticSessions::from_path)
+        .transpose()
+        .unwrap();
```
davehadley: expect is preferred over unwrap

charts/sessionspaces/staging-values.yaml
```diff
@@ -0,0 +1,11 @@
+staticSessions:
+  enabled: true
+  sessions:
+    - proposal_code: ks
+      proposal_number: 10000
+      visits: [1, 2, 3, 4, 5]
```
davehadley, line 6: why 5?

charts/sessionspaces/staging-values.yaml
```diff
@@ -0,0 +1,11 @@
...
+      instrument: t01
+      members: []
+      gid: null
```
davehadley: is it OK to have a null gid?

charts/sessionspaces/staging-values.yaml
```diff
@@ -0,0 +1,11 @@
+staticSessions:
+  enabled: true
+  sessions:
+    - proposal_code: ks
+      proposal_number: 10000
+      visits: [1, 2, 3, 4, 5]
+      instrument: t01
+      members: []
```
davehadley, line 8: a session with no members is not very useful?

davehadley (approved):

charts/sessionspaces/staging-values.yaml
TBThomas56: Agreed in team for users to add their own information to the configmap so that there is no sensitive info on the repo

charts/sessionspaces/staging-values.yaml
TBThomas56: One for every permanent member of the team

charts/sessionspaces/staging-values.yaml
TBThomas56: updated - good catch

backend/sessionspaces/src/main.rs
TBThomas56: after the statics.merge_into line, update_sessionspaces is called and the it checks if there are any changes between current_session and new_session. Since this is not there, there is no application - same functionality with other sessionspaces

PR #1284, opened 2026-05-06
feat(charts): implement metacontroller for events

JamesDoingStuff:
[AP-1095](https://jira.diamond.ac.uk/browse/AP-1095)


charts/events/hooks/sync.py
```diff
@@ -0,0 +1,154 @@
+from http.server import BaseHTTPRequestHandler, HTTPServer
+import json
+
+class Controller(BaseHTTPRequestHandler):
+  def sync(self, parent: dict, related: dict):
```
davehadley: Can we add type hints to this python code?

davehadley (approved):

charts/events/hooks/sync.py
TBThomas56: Can do that with pydantic if I am not mistaken

charts/events/values.yaml
TBThomas56, line 1: Anything different between staging and values?

JamesDoingStuff: Not currently - `events` is already enabled on prod, all this does is add the metacontroller and currently no-one outside the team can create EventSources/Triggers

charts/events/hooks/sync.py
JamesDoingStuff: I've made the typing a bit more thorough (as much as possible without writing up full k8s schemas, anyway). I haven't implemented pydantic yet as I don't want to enforce anything while I'm still working out exactly what the API will look like

davehadley (approved):

PR #1294, opened 2026-05-11
Fix for the workflow cli  create cmd

hazdl:
Fixed the issues in the cli create cmd. Under the issue :AP:1099.


backend/auth-daemon/README.md
```diff
@@ -1,6 +1,6 @@
 # auth-daemon
 
-Sidecar HTTP proxy that injects valid OIDC access tokens into GraphQL requests made from within Argo workflows. Workflows call the daemon on `localhost` instead of the GraphQL API directly; the daemon handles token refresh and auth header injection transparently.
+ Sidecar HTTP proxy that injects valid OIDC access tokens into GraphQL requests made from within Argo workflows. Workflows call the daemon on `localhost` instead of the GraphQL API directly; the daemon handles token refresh and auth header injection transparently.
```
davehadley, line 3: this edit seems un-related to this pull request?

workflows-cli/src/create/mod.rs
```diff
@@ -59,61 +87,69 @@ fn prompt(message: &str) -> bool {
...
+
+        let metadata = fs::symlink_metadata(&src_path)
+            .map_err(|e| format!("Failed to read metadata for {}: {}", src_path.display(), e))?;
+
+        // SYMLINK
+        if metadata.file_type().is_symlink() {
+            let target = fs::read_link(&src_path)
+                .map_err(|e| format!("Failed to read symlink {}: {}", src_path.display(), e))?;
+
+            fs_sym::symlink(&target, &dest_path)
```
davehadley, line 130: you seem to have lost #[cfg(unix)]?

workflows-cli/src/create/mod.rs
```diff
@@ -19,38 +19,66 @@ pub fn create(args: CreateArgs) {
 }
 
 fn generate_template_repo(args: &CreateArgs, prompt_fn: fn(&str) -> bool) -> Result<(), String> {
```
davehadley, line 21: There's a lot of change here, some of it seems to be refactoring.

What actually was the cause of the bug? And what was the fix?
Is it possible to create a unit test for this bug to prevent future regressions?

examples/conventional-templates/hello_cli_test.yaml
```diff
@@ -0,0 +1,26 @@
+
```
davehadley, line 1: Is this necessary for this PR? I don't think that it shouldn't be included in the `examples/conventional-templates` directory. 

1. This is not a `ClusterWorkflowTemplate`, it is a `Workflow`. These are distinct concepts:
    - <https://argo-workflows.readthedocs.io/en/latest/cluster-workflow-templates/> 
    - <https://argo-workflows.readthedocs.io/en/latest/workflow-concepts/>
2. Templates in the `examples` directory are user facing. That will get ingested by ArgoCD and appear at <https://workflows.diamond.ac.uk> where users and see it and run it.
3. It would be better to have an automated rust unit test to test CLI functionality. Not a Workflow that you have to manually submit to the cluster,

davehadley (changes requested):
Thanks @hazdl for implementing this fix.

There seems to be refactoring in the PR as well a bug fix. Where possible, it's easier to review if you do the minimal changes required to fix the bug, and the refactor as a separate PR.

In the workflows repo we follow conventional commits. See: <https://www.conventionalcommits.org/en/v1.0.0/>. And we normally merge PRs without squashing (i.e. the PR commit history will get linearly merged into main).

I suggest squashing this PRs commits into 1 or 2 commits: eg 
"fix(workflows-cli): summary of bug fixed"
"refactor(workflows-cli): summary of refactoring done"

Thanks again for fixing this! 👍

PR #1297, opened 2026-05-12
Postgres Backup & Restore

JamesDoingStuff:
[AP-1021](https://jira.diamond.ac.uk/browse/AP-1021)
- Replaces the old rclone secret, fixing the etcd backup.
- Adds a CronJob to automatically backup the Postgres database and a script that can be used to restore it. This is enabled for Pollux, but currently disabled for Argus

I've tested that the etcd restore can replace deleted configmaps and the Postgres one can bring back workflows after deleting the database volumes. The etcd restoration leaves Argo CD unable to start and breaks the Argo Workflows PVCs. This is less than ideal, but the postgres backup will at least prevent complete data loss - the complete restore process is documented [here](https://gitlab.diamond.ac.uk/analysis-platform/analysis-platform-internal-docs/-/merge_requests/2). I propose merging this as-is so that the automated backups can begin, then we can improve the restore procedure in future


charts/workflows-cluster/templates/postgres-restore-job.yaml
```diff
@@ -0,0 +1,77 @@
+{{- if .Values.runPostgresRestore }}
```
davehadley, line 1: What is the intended usage here?

I guess someone does something like:
```
helm upgrade --install workflows-cluster charts/workflows-cluster/ -n workflows --set runPostgresRestore=true
```

If someone later does:
```
helm upgrade --install workflows-cluster charts/workflows-cluster/ -n workflows --reuse-values
```
does the restore job get re-run?

davehadley (approved):

charts/workflows-cluster/templates/postgres-restore-job.yaml
JamesDoingStuff: It's a bit of a workaround. I needed a way to be able to create a Job inside the vcluster as part of the script, but because I'd templated some parts of it, I had to leave it in `templates`. The problem was, whenever a `helm upgrade` ran, the Job was created and immediately tried to restore the database. My solution was having a Value that is always non-existent unless manually passed in, which happens in this line of the restore script:
```
helm template . -s templates/postgres-restore-job.yaml --set runPostgresRestore=true | kubectl -n workflows apply -f -
```
If someone were to run that upgrade command then yes, it would get re-run the next time, but it's a value that needn't be set outside of the script. Do you think I should put a comment on the template to that effect?

PR #1306, opened 2026-05-14
Logout Button

JamesDoingStuff:
[AP-1079](https://jira.diamond.ac.uk/browse/AP-1079)
Adds logout button to the navbar that redirects to `https://identity.diamond.ac.uk/realms/dls/protocol/openid-connect/logout`.

### DIFF
--- a/frontend/dashboard/src/RelayEnvironment.ts
+++ b/frontend/dashboard/src/RelayEnvironment.ts
@@ -10,6 +10,9 @@ import {
 } from "relay-runtime";
 import { getKeycloak } from "./keycloak";
 import { createClient } from "graphql-ws";
+import { AuthState } from "@diamondlightsource/sci-react-ui";
+import { parseJwt } from "./routes/utils";
+import { JSONObject } from "workflows-lib";
 
 const HTTP_ENDPOINT = import.meta.env.VITE_GRAPH_URL;
 const WS_ENDPOINT = import.meta.env.VITE_GRAPH_WS_URL;
@@ -55,7 +58,6 @@ const fetchFn: FetchFunction = async (request, variables) => {
   if (!keycloak.authenticated) {
     await ensureKeycloakInit();
   }
-
   if (keycloak.token) {
     const resp = await fetch(HTTP_ENDPOINT, {
       method: "POST",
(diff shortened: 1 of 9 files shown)


frontend/dashboard/src/RelayEnvironment.ts
```diff
@@ -130,3 +131,17 @@ export async function getRelayEnvironment(): Promise<Environment> {
   }
   return RelayEnvironment;
 }
+
+export async function getUser(): Promise<AuthState | null> {
+  if (!keycloak.authenticated) {
+    await ensureKeycloakInit();
+  }
+  if (keycloak.token) {
+    const parsedToken = parseJwt(keycloak.token);
```
davehadley: can `parseJwt` fail and raise an exception? if so, does that need to be handled?

frontend/dashboard/src/mocks/mockKeycloak.ts
```diff
@@ -4,14 +4,17 @@ const createMockJwt = (payload: object) => {
...
-const mockToken = createMockJwt({
+export const mockJwtPayload = {
   sub: "mock-user",
   preferred_username: "mockuser",
   email: "mock@diamond.ac.uk",
   exp: Number.MAX_SAFE_INTEGER,
   iat: Number.MIN_SAFE_INTEGER,
   iss: "https://authn.diamond.ac.uk/realms/master",
-});
+  name: "Mo C. Kuser",
```
davehadley, line 14: nice name!

frontend/dashboard/src/RelayEnvironment.ts
```diff
@@ -130,3 +131,17 @@ export async function getRelayEnvironment(): Promise<Environment> {
...
+    const user: AuthState = {
+      name: parsedToken.name as string,
+      fedid: (parsedToken.preferred_username ?? parsedToken.fedid) as string,
```
davehadley, line 149: I think that not all user accounts are guaranteed to have a fedid. I don't know if that needs to be considered.

davehadley (approved):

frontend/dashboard/src/RelayEnvironment.ts
JamesDoingStuff: This will leave the fedid as `undefined` in that case, which just means there isn't a username displayed in the user profile component. Added a test case for this

frontend/dashboard/src/RelayEnvironment.ts
JamesDoingStuff: Added a catch for this :+1:

PR #1322, opened 2026-05-19
Standardise CPU and Memory Resources for Argo Workflows

hazdl:
This PR standardises CPU and memory requests/limits across the Argo Workflows Helm chart to  ensure consistent resource allocation for controller, server, and workflow execution pods.

-Updated controller resources to:   requests: 500m CPU / 2Gi memory
                                                       limits: 1 CPU / 4Gi memory
-Updated server resources to:   requests: 1 CPU / 1Gi memory
                                                   limits: 2 CPU / 2Gi memory
-Added/updated workflowDefaults.spec.podSpecPatch to enforce workflow pod resources:
                                                  requests: 500m CPU / 2Gi memory
                                                  limits: 1 CPU / 4Gi memory
-Applies to:     new workflow executions only
                      No changes required to existing workflows

Verification:
Verified via helm template rendering that workflowDefaults.spec.podSpecPatch applies CPU and memory requests/limits to the workflow main container. Confirmed in rendered output:
workflowDefaults.spec.podSpecPatch → containers.main.resources

Confirmed using Helm-rendered manifests and grep validation.
grep -n "podSpecPatch" rendered.yaml
grep -n "2Gi" rendered.yaml
grep -n "500m" rendered.yaml

### DIFF
--- a/charts/workflows/values.yaml
+++ b/charts/workflows/values.yaml
@@ -1,5 +1,6 @@
 argo-workflows:
   enabled: true
+
   artifactRepository:
     archiveLogs: true
     s3:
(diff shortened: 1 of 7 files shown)


charts/workflows/values.yaml
```diff
@@ -1,5 +1,6 @@
 argo-workflows:
   enabled: true
+
```
davehadley, line 3: There is a lot of formatting change noise in this PR.

This makes it hard to review and creates noise in diffs. Please try not to do this where possible.

charts/workflows/values.yaml
```diff
@@ -12,128 +13,192 @@ argo-workflows:
...
           name: postgres-argo-workflows-password
           key: username
+
         passwordSecret:
           name: postgres-argo-workflows-password
           key: password
+
     workflowDefaults:
       spec:
+        podSpecPatch: |
```
davehadley: I think that this modifies the default Workflow pods resources? What is the motivation for this?

I don't think this is needed for [AP-1119](https://jira.diamond.ac.uk/browse/AP-1119)?

charts/workflows/values.yaml
```diff
@@ -12,128 +13,192 @@ argo-workflows:
...
+                      operator: In
+                      values:
+                        - workflows
+
     extraEnv:
       - name: POD_NAMES
         value: "v1"
+
   server:
+    resources:
```
davehadley, line 91: This change halves the server pod memory and doubles the cpu compared to what is currently the case on argus:

```
      limits:
        cpu: "1"
        memory: 4Gi
      requests:
        cpu: 500m
        memory: 2Gi
```


Is this intentional?

charts/workflows/values.yaml
```diff
@@ -147,58 +212,73 @@ oauth2-proxy:
...
             path: secret
+
   extraVolumeMounts:
     - name: secret
       mountPath: /etc/alpha
       readOnly: true
 
 s3sealedsecret:
   enabled: true
+
```
davehadley, line 270: The commit history should be cleaned up (it would be better to combine these 4 commits into one).

davehadley (changes requested):

charts/workflows/values.yaml
hazdl: Yes — this was based on the issue description and log,  which identified workloads already exceeding ~1Gi memory usage and recommended explicitly setting requests/limits (e.g. ~2Gi/4Gi) instead of relying on defaults. I aligned the values accordingly for consistency and stability. Happy to adjust if needed to match argus baseline exactly.

charts/workflows/values.yaml
hazdl: You are correct — this change goes beyond AP-1119.
The intention was to align workflow execution pods with the same resource defaults, but this is not strictly required for this ticket.

I can remove workflowDefaults.spec.podSpecPatch from this PR and limit the change to controller/server resources only if preferred, and raise a separate PR for workflow pod defaults.

charts/workflows/values.yaml
hazdl: Had issues with the spacing, so edit externally and paste in vim. I guess this was the reason.
Will avoid in the future

charts/workflows/values.yaml
hazdl: I’ve rebased the branch and squashed the 4 commits into a single commit to improve commit history clarity.

charts/workflows/values.yaml
davehadley: Setting default limits on pods may be a good idea but we should discuss what they should be before implementing.

davehadley (approved):

PR #1327, opened 2026-05-20
fix(charts): fix graph-proxy-k6-auth sync issues

TBThomas56:


charts/workflows-cluster/charts/secrets/templates/graph-proxy-k6-auth.yaml
```diff
@@ -2,8 +2,13 @@
 apiVersion: bitnami.com/v1alpha1
 kind: SealedSecret
 metadata:
-  name: graph-proxy-k6-auth  
+  name: graph-proxy-k6-auth
   namespace: workflows
+  labels:
```
davehadley, line 7: I don't understand why this is needed? this isn't needed on any of the other secrets?

charts/workflows-cluster/charts/secrets/templates/graph-proxy-k6-auth.yaml
```diff
@@ -17,8 +22,13 @@ spec:
...
```
davehadley, line 27: same coment as above?

davehadley (approved):

PR #1329, opened 2026-05-20
fix(charts): fix indenting in cronJob test as that caused job to fail

MahirAbbas:


davehadley (approved):

charts/workflows-cluster/Chart.yaml
```diff
@@ -3,7 +3,7 @@ name: workflows-cluster
 description: A virtual cluster for Data Analysis workflows
 type: application
 
-version: 0.13.6
+version: 0.13.7
```
TBThomas56: do we need to upgrade the chart here or just the k6operator

TBThomas56 (changes requested):

charts/workflows-cluster/Chart.yaml
MahirAbbas: CI fails without updating the chart.

TBThomas56 (approved):

PR #1333, opened 2026-05-22
Tbt/auth daemon config secret

TBThomas56:
Distributes auth-daemon-config secret to session namespaces from workflows namespace (GeneratingPolicy) and enforces access controls to protect sensitive content

Cronjob acts as fallback for existing namespaces and secret rotation.

Kyverno clusterpolicy enforces that only containers running the auth-daemon image may mount or reference the secret and blocks any exec into any pod that runs auth-daemon. 

Tested and works on pollux. Must be merged after #1340


davehadley (approved):
Yikes! More state being synced to visit namespaces. 😭

PR #1340, opened 2026-05-26
Tbt/auth daemon patch

TBThomas56:


backend/auth-daemon/src/main.rs
```diff
@@ -27,9 +28,50 @@ mod state;
...
 use config::DaemonConfig;
+use kube::{
+    Api, Client,
+    api::{ApiResource, DynamicObject},
+    core::GroupVersionKind,
+};
 
 type Result<T> = std::result::Result<T, Error>;
 
+static CRYPTO_PROVIDER: OnceLock<()> = OnceLock::new();
```
davehadley, line 39: You refactored it to make a kubernetes API call to get the subject from the Workflow rather than taking it from the env var WORKFLOWS_AUTH_DAEMON_SUBJECT.

Why was this refactor needed?

If it's for security reasons, is this more secure? If a malicious user can manipulate the env var WORKFLOWS_AUTH_DAEMON_SUBJECT then can they not also manipulate ARGO_WORKFLOW_NAME?

davehadley (approved):

backend/auth-daemon/src/main.rs
TBThomas56: the subject is the Keycloak subject for each user. We do not want users having to add their subject as a parameter and it can be automatically received from th K8s API. Definitely more secure

Argo_workflow_name changes with the random string added at the end. So we want to get it from the workflow as we cannot know from before

TBThomas56: Discussed and agreed to make change in future commit for a working copy to be present

PR #1350, opened 2026-05-28
feat(charts): add stomp event server

JamesDoingStuff:
The main addition in this PR is a '[generic' EventSource](https://argoproj.github.io/argo-events/eventsources/generic/) for handling STOMP messages. A generic EventSource requires a server that implements the gRPC contract at the previous link. It is set up using the `config` field in the EventSource - this can contain anything, provided the server can parse it. 

My server takes in the stomp connection config, creates the connection and then adds any received 'scan start' messages to a queue that gets turned into Events.  Only the 'start' messages contain the visitId, so for now only those are being processed. Eventually, subsequent messages will need to be matched to their scan start message by the scan uid. I also intend to move the filtering logic [into the EventSource](https://argoproj.github.io/argo-events/eventsources/filtering/) itself in future, but I haven't had a chance to test that yet. 

I have put the server code in a ConfigMap that gets ran in a plain python container - this could be replaced by a custom server container down the road. It also remains to be seen whether or not this server can be shared between multiple EventSources

### DIFF
--- a/charts/events/generic-servers/stomp-server.py
+++ b/charts/events/generic-servers/stomp-server.py
@@ -1,0 +1,108 @@
+import generic_pb2
+import generic_pb2_grpc
+import stomp
+import grpc
+import logging
+import stomp.utils
+import queue
+import json
+import os
+from concurrent import futures
+from pydantic import BaseModel, ConfigDict
+from typing import Literal
+
+
+class EventSourceConfig(BaseModel):
+    model_config = ConfigDict(validate_assignment=True)
+
+    user: str
+    password: str
+    host: str
+    port: int
+    destination: str
+    ack: Literal["auto", "client", "client-individual"] = "auto"
+
+def parse_config(config: bytes) -> EventSourceConfig | None:
+    try:
+        decoded_confid = config.decode()
+        config_entries = decoded_confid.strip().split("\n")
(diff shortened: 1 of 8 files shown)


charts/events/generic-servers/stomp-server.py
```diff
@@ -0,0 +1,107 @@
...
+    generic_pb2_grpc.add_EventingServicer_to_server(StompEventServicer(), server)
+    server.add_insecure_port("[::]:" + port)
+    server.start()
+    logging.info("Server started, listening on " + port)
+    server.wait_for_termination()
+
+
+if __name__ == "__main__":
+    logging.basicConfig(
+        level=logging.DEBUG, format="%(asctime)s %(levelname)s %(name)s - %(message)s"
```
davehadley: what is DEBUG log volume? do these pods log to graylog (all pods do by default), cloud team don't like it if the graylog log volume is too high. consider setting the default to INFO

davehadley (approved):

charts/events/generic-servers/stomp-server.py
JamesDoingStuff: It was logging heartbeats, so probably more than necessary. I've set it to INFO by default, with the option to set it in `Values`

PR #1364, opened 2026-06-02
feat(charts): allow auto workflows to set own posixuid

JamesDoingStuff:
[AP-1110](https://jira.diamond.ac.uk/browse/AP-1110)

I had to remove the pipe expressions from the JMESPath in the policy - they don't seem to work with the || operator. Possibly a kyverno bug, as they seem to behave as one would expect in the JMESPath tutorial/playground.

If `identity-mapper` changes related to [AP-1140](https://jira.diamond.ac.uk/browse/AP-1140) are merged in before this, then that policy will need updating instead in a similar way to how `workflow-label-clusterpolicy` is here e.g. 
```
context:
     - name: posixUidString
       variable:
         value: "{{`{{ request.userInfo.extra.\"workflows.diamond.ac.uk/posixuid\"[0] || request.object.metadata.labels.\"workflows.diamond.ac.uk/machine-uid\" }}`}}"
```

### DIFF
--- a/charts/workflows/templates/workflow-label-clusterpolicy.yaml
+++ b/charts/workflows/templates/workflow-label-clusterpolicy.yaml
@@ -13,6 +13,10 @@ spec:
             - argoproj.io/*/Workflow
           operations:
             - CREATE
+      context:
+        - name: machineuid
+          variable:
+            jmesPath: request.object.metadata.labels."workflows.diamond.ac.uk/machine-uid" || ""
       mutate:
         patchStrategicMerge:
           metadata:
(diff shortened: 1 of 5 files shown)


charts/workflows/templates/workflow-label-clusterpolicy.yaml
```diff
@@ -13,12 +13,16 @@ spec:
...
+            jmesPath: request.object.metadata.labels."workflows.diamond.ac.uk/machine-uid" || ""
       mutate:
         patchStrategicMerge:
           metadata:
             labels:
               {{- with .Values.uid }}
               workflows.diamond.ac.uk/creator-posix-uid: '{{ . }}'
               {{- else }}
-              workflows.diamond.ac.uk/creator-posix-uid: '{{ `{{ request.userInfo.extra | "workflows.diamond.ac.uk/posixuid" | [0] }}` }}'
+              workflows.diamond.ac.uk/creator-posix-uid: '{{ `{{ request.userInfo.extra."workflows.diamond.ac.uk/posixuid"[0] || machineuid }}` }}'
```
davehadley, line 27: checking I understand the safety reasoning here: it's impossible for a malicious user to set "workflows.diamond.ac.uk/machine-uid" themselves because "workflows.diamond.ac.uk/posixuid" is always set and is out of their control?

davehadley (approved):

charts/workflows/templates/workflow-label-clusterpolicy.yaml
JamesDoingStuff: Yes that was my thinking - I just tried running a workflow with the label set manually and the pod was still my UID rather than the machine one

TBThomas56 (approved):

PR #1366, opened 2026-06-02
feat(identity-mapper): MVP tool to apply user groups compliant with incoming cluster admin policies

davehadley:


JamesDoingStuff (approved):

.github/workflows/ci.yaml
```diff
@@ -5,6 +5,22 @@ on:
   pull_request:
 
 jobs:
+
+  changes:
```
TBThomas56: should there be a better name for this ci?
My understanding is that it is "changes"

charts/identity-mapper/templates/cronjob.yaml
TBThomas56, line 1: Fine for MVP but this is a lot of memory being used at once to create identity for every person at Diamond. Maybe implementing some form of pagination / multiple K8s API calls could be done?

backend/identity-mapper/src/identity_mapper/_lookup_identities_in_ldap.py
```diff
@@ -0,0 +1,81 @@
...
+
+from ._identity import Identity
+
+_BASE_DN = "dc=diamond,dc=ac,dc=uk"
+
+
+def lookup_identities_in_ldap(
+    ldap: ldap3.Connection,
+) -> dict[int, Identity]:
+    people_base_dn = _BASE_DN
```
TBThomas56, line 11: the key here is int. but in sync it is a string. Is that on purpose?  Wouldn't that cause the difference in the K8s state always? And user identity to be recreated every 30 minutes?

backend/identity-mapper/src/identity_mapper/_sync_ldap_to_kubernetes.py
```diff
@@ -0,0 +1,71 @@
...
+_logger = logging.getLogger(__name__)
+
+
+def sync_ldap_to_kubernetes(
+    kubectl: kubernetes.client.CustomObjectsApi,
+    ldap_data: dict[int, Identity],
+    kubernetes_data: dict[int, Identity],
+) -> None:
+    desired_state = {
+        str(uid_num): {
```
TBThomas56: referencing to different key type from above

backend/identity-mapper/src/identity_mapper/_sync_ldap_to_kubernetes.py
```diff
@@ -0,0 +1,71 @@
...
+        "Applying %d updates to match desired LDAP state...",
+        len(patch.patch),
+    )
+
+    for operation in patch.patch:
+        op_type = operation["op"]  # 'add', 'remove', or 'replace'
+        path_parts = operation["path"].strip("/").split("/")
+        username = path_parts[0]  # The user ID string (e.g., "10023")
+
+        if op_type == "add" and len(path_parts) == 1:
```
TBThomas56: for next review. Probably better to use a match -> case instead of if / elif

backend/identity-mapper/src/identity_mapper/_sync_ldap_to_kubernetes.py
```diff
@@ -0,0 +1,71 @@
...
+                "apiVersion": f"{IdentityCrd.GROUP}/{IdentityCrd.VERSION}",
+                "kind": "UserIdentity",
+                "metadata": {"name": username},
+                "spec": operation["value"],
+            }
+            kubectl.create_cluster_custom_object(
+                IdentityCrd.GROUP, IdentityCrd.VERSION, IdentityCrd.PLURAL, body
+            )
+
+        elif op_type == "remove" and len(path_parts) == 1:
```
TBThomas56: what happens if you try remove a suplementalgroup? {"op": "remove", "path": "/1001/suplementalGroups/dasc"}. How is this handled as the len of path_parts are 3 here

backend/identity-mapper/tests/test_kubernetes.py
```diff
@@ -0,0 +1,57 @@
...
+
+import kubernetes.client
+
+from identity_mapper._identity import IdentityCrd
+from identity_mapper._lookup_identities_in_kubernetes import (
+    lookup_identities_in_kubernetes,
+)
+
+
+def test_lookup_identities_in_kubernetes() -> None:
```
TBThomas56, line 11: could add a test case of syncing / updating to ensure that the things are not being deleted and brought back

backend/identity-mapper/pyproject.toml
```diff
@@ -0,0 +1,35 @@
...
+description = "Determines pod security context for Diamond Workflows"
+readme = "README.md"
+authors = [
+    { name = "David Hadley", email = "davehadley@users.noreply.github.com" }
+]
+requires-python = ">=3.13"
+dependencies = [
+    "kubernetes==35.0.0",
+    "ldap3>=2.9.1",
+    "structlog>=25.5.0",
```
TBThomas56: do not see structlog being used but logging is used throughout.

charts/identity-mapper/templates/policy.yaml
```diff
@@ -0,0 +1,149 @@
...
+
+          - list: "request.object.spec.ephemeralContainers || `[]`"
+            patchStrategicMerge:
+              spec:
+                ephemeralContainers:
+                  - (name): "{{`{{ element.name }}`}}"
+                    securityContext:
+                      runAsUser: "{{`{{ request.object.spec.securityContext.runAsUser }}`}}"
+                      runAsGroup: "{{`{{ request.object.spec.securityContext.runAsGroup }}`}}"
+{{- end }}
```
TBThomas56: no new line

charts/identity-mapper/test-policy/mutate-argo-workflow-security-context/chainsaw-test.yaml
```diff
@@ -0,0 +1,137 @@
...
+#                       runAsUser: 4321
+#                       allowPrivilegeEscalation: false
+#                       readOnlyRootFilesystem: true
+apiVersion: chainsaw.kyverno.io/v1alpha1
+kind: Test
+metadata:
+  name: noop
+spec:
+  steps:
+    - try:
```
TBThomas56: no new line

charts/apps/staging-values.yaml
```diff
@@ -92,3 +92,10 @@ k6Operator:
   extraValueFiles:
     - staging-values.yaml
   valuesObject: {}
+
+identityMapper:
+  enabled: true
+  targetRevision: HEAD
+  extraValueFiles:
+    - staging-values.yaml
+  valuesObject: {}
```
TBThomas56, line 101: no new line

charts/apps/dev-values.yaml
```diff
@@ -70,3 +70,9 @@ dashboard:
   targetRevision: HEAD
   extraValueFiles:
     - dev-values.yaml
+
+identityMapper:
+  enabled: false
+  targetRevision: HEAD
+  extraValueFiles:
+    - dev-values.yaml
```
TBThomas56: no new line

backend/identity-mapper/Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+COPY . /app
+
+# Disable development dependencies
+ENV UV_NO_DEV=1
+
+# Sync the project into a new environment, asserting the lockfile is up to date
+WORKDIR /app
+RUN uv sync --locked
+
+CMD ["uv", "run", "--no-sync", "identity-mapper"]
```
TBThomas56: no new line

TBThomas56 (changes requested):

.github/workflows/ci.yaml
davehadley: added clearer name

charts/identity-mapper/templates/cronjob.yaml
davehadley: The peak memory usage of the stored LDAP + Kubernetes information are ~700 MB.
When converted to rust this will likely be an order of magnitude lower.

backend/identity-mapper/src/identity_mapper/_lookup_identities_in_ldap.py
davehadley: `sync_ldap_to_kubernetes` expected `int` keys:
```
def sync_ldap_to_kubernetes(
    kubectl: kubernetes.client.CustomObjectsApi,
    ldap_data: dict[int, Identity],
    kubernetes_data: dict[int, Identity],
)
```

davehadley: Identities are (currently) *synced* every 30 minutes.

They are only applied if they have changed.

This period will be changed in a future PR to match the period of cloud teams updates (3 hours).

TBThomas56: okay. I understood that. Just a check for when updating, will it be seen 'as a change' if you poll integers (as required by ldap) and then cross-referencing it with `desired_state` in _sync_ldap_to_k8s.py which has it as a str(uid_name)

davehadley: I see your point about int vs string for the key. jsonpatch seems not care about this. I think because json object keys are required to be strings. jsonpatch is probably internally converting to string.

However, I am now treating both dicts consistently.

backend/identity-mapper/src/identity_mapper/_sync_ldap_to_kubernetes.py
davehadley: see comments above

backend/identity-mapper/src/identity_mapper/_sync_ldap_to_kubernetes.py
davehadley: ~~done~~

refactored so no longe relevant

backend/identity-mapper/src/identity_mapper/_lookup_identities_in_ldap.py
davehadley: Actually I noticed Json patch can produce multiple patches per object resulting in multiple API calls per object. So I have removed the json patch dependency and manually calculate the updates.

backend/identity-mapper/src/identity_mapper/_sync_ldap_to_kubernetes.py
davehadley: I have added a test for removing a supplementalGroup.

Implementation has been changed, path_parts comment is no longer relevant.

backend/identity-mapper/tests/test_kubernetes.py
davehadley: I don't understand this comment. This file contains unit tests for `lookup_identities_in_kubernetes`. `lookup_identities_in_kubernetes` is a read only operation.

backend/identity-mapper/pyproject.toml
davehadley: this is from a version where struct logging was used. removed.

charts/identity-mapper/templates/policy.yaml
davehadley: fixed.

would be nice to one day find a decent helm auto-formatter to have these things fixed automatically and checked in CI.

charts/identity-mapper/test-policy/mutate-argo-workflow-security-context/chainsaw-test.yaml
davehadley: done.

charts/apps/staging-values.yaml
davehadley: done

charts/apps/dev-values.yaml
davehadley: done

backend/identity-mapper/Dockerfile
davehadley: done.

backend/identity-mapper/tests/test_kubernetes.py
TBThomas56: Wanted to check if there were multiple api calls per object which was handled

TBThomas56 (approved):

backend/identity-mapper/tests/test_kubernetes.py
davehadley: yep the sync tests already check for that

davehadley: and you were correct, there were multiple api calls in the original code

PR #1373, opened 2026-06-08
fix(graph-proxy): subscriptions enforces authentication with AuthGuard

davehadley:
Fixes a bug in graph-proxy that broke GraphQL subscriptions and produced: `"Authentication context missing"` errors visible in the console on the frontend.

This PR fixes bug and adds unit tests to prevent future regression.

### DIFF
--- a/backend/graph-proxy/src/graphql/subscription_integration.rs
+++ b/backend/graph-proxy/src/graphql/subscription_integration.rs
@@ -14,7 +14,7 @@
 // accessible inside the graph functions - after upgrading
 // requests to websocket, which is not natively supported.
 
-use std::{convert::Infallible, future::Future, str::FromStr, time::Duration};
+use std::{convert::Infallible, future::Future, str::FromStr, sync::Arc, time::Duration};
 
 use async_graphql::{
     futures_util::task::{Context, Poll},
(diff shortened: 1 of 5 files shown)


backend/graph-proxy/src/graphql/subscription_integration.rs
```diff
@@ -457,20 +488,25 @@ mod tests {
...
+
+        let expected_token = access_token.token().to_string();
 
         let mut req = server_address.into_client_request()?;
-        req.headers_mut().insert(
-            http::header::AUTHORIZATION,
-            http::HeaderValue::from_str(&format!("Bearer {token}"))?,
-        );
+        req.headers_mut().typed_insert(access_token);
+        //req.headers_mut().insert(http::header::AUTHORIZATION, ???);
```
JamesDoingStuff: Can this comment be removed, or is this something that needs to be re-implemented in future?

JamesDoingStuff (approved):

backend/graph-proxy/src/graphql/subscription_integration.rs
davehadley: that was left in by accident. removed

PR #1375, opened 2026-06-08
chore(charts): switch on identity-mapper on staging

davehadley:
Needs to be merged after #1374

### DIFF
--- a/charts/sessionspaces/staging-values.yaml
+++ b/charts/sessionspaces/staging-values.yaml
@@ -10,3 +10,6 @@ staticSessions:
       gid: 123456
       start_date: "2024-01-01 00:00:00"
       end_date: "2099-12-31 23:59:59"
+
+policy:
+  create: false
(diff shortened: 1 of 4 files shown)


charts/sessionspaces/staging-values.yaml
```diff
@@ -10,3 +10,6 @@ staticSessions:
       gid: 123456
       start_date: "2024-01-01 00:00:00"
       end_date: "2099-12-31 23:59:59"
+
+policy:
+  create: false
```
JamesDoingStuff: Missing newline

JamesDoingStuff (commented):
Question that I probably should have asked on #1366 - with this new policy, what happens if I manage to submit a workflow without a `workflows.diamond.ac.uk/posixuid`? Previously they'd fail only because the path couldn't be resolved, but I can't immediately work out what will happen here, as the policy preconditions looks like they'd prevent the policy from being applied at all

JamesDoingStuff (approved):

charts/sessionspaces/staging-values.yaml
davehadley: fixed

PR #1376, opened 2026-06-08
chore(frontend): isolate graphql components

JamesDoingStuff:
[AP-1142](https://jira.diamond.ac.uk/browse/AP-1142)

- Moves `mocks`, `RelayEnvironment`, `keycloak.ts` out of dashboard
- Moves `WorkflowsNavbar` out of workflows-lib
- Remove `relay-compiler`, `react-relay`  and `relay-types` as dependencies from dashboard. I think the babel/vite plugins have to stay in the dashboard, as that it where the relay transpilation actually occurs - if there's a way of changing this, do let me know


davehadley (approved):
LGTM

PR #1377, opened 2026-06-08
fix(template-linter): filter out temporary files from lint results

hazdl:
- Temporary files generated during linting were previously included in results, causing noise in CI and local lint outputs. This ensures only meaningful files are considered.
- Verified CI passes successfully after the change
- passed all CI checks including formatting, clippy, and build tests

### DIFF
--- a/workflows-cli/src/linter/helm.rs
+++ b/workflows-cli/src/linter/helm.rs
@@ -17,7 +17,12 @@ pub fn lint_from_helm(target: &Path, all: bool) -> Result<Vec<LintResult>, Strin
         .map_err(|_e| "Couldn't create temporary file for helm templates.")?;
 
     let path_buf = tmp_dir.to_path_buf();
-    lint_from_manifest(&path_buf, true)
+    let mut results = lint_from_manifest(&path_buf, true)?;
+
+    // Filter out tmp files
+    results.retain(|r| !r.template_name.starts_with("/tmp/argo-lint"));
+
+    Ok(results)
 }
 
 pub fn write_to_clean_folder(path: &Path, contents: Vec<String>) -> std::io::Result<()> {
--- a/workflows-cli/src/linter/helm.rs
+++ b/workflows-cli/src/linter/helm.rs
@@ -20,7 +20,7 @@ pub fn lint_from_helm(target: &Path, all: bool) -> Result<Vec<LintResult>, Strin
     let mut results = lint_from_manifest(&path_buf, true)?;
 
     // Filter out tmp files
-    results.retain(|r| !r.template_name.starts_with("/tmp/argo-lint"));
+    results.retain(|r| !r.name.starts_with("/tmp/argo-lint"));
 
     Ok(results)
 }


workflows-cli/src/linter/helm.rs
```diff
@@ -17,7 +17,12 @@ pub fn lint_from_helm(target: &Path, all: bool) -> Result<Vec<LintResult>, Strin
         .map_err(|_e| "Couldn't create temporary file for helm templates.")?;
 
     let path_buf = tmp_dir.to_path_buf();
-    lint_from_manifest(&path_buf, true)
+    let mut results = lint_from_manifest(&path_buf, true)?;
+
+    // Filter out tmp files
+    results.retain(|r| !r.name.starts_with("/tmp/argo-lint"));
```
davehadley, line 23: Thanks for working on this. Maybe I don't understand but what problem does this actually solve?

This is the current implementation:

```rust
pub fn lint_from_helm(target: &Path, all: bool) -> Result<Vec<LintResult>, String> {
    // FIXME: does not respect $TMPDIR and potential race condition / permissions issue on multi-user systems
    let tmp_dir = Path::new("/tmp/argo-lint");
    let manifests = helm_to_manifest(target, all)?;
    write_to_clean_folder(tmp_dir, manifests)
        .map_err(|_e| "Couldn't create temporary file for helm templates.")?;

    let path_buf = tmp_dir.to_path_buf();
    lint_from_manifest(&path_buf, true)
}
```

As far as I can tell this:

(1) Runs helm on helm templates to produce manifests.
(2) Writes manifests to `/tmp/argo-lint`
(3) Runs argo linter on the files in `/tmp/argo-lint`.

If you then do:

```
results.retain(|r| !r.name.starts_with("/tmp/argo-lint"));
```

you remove all results from files in `/tmp/argo-lint` which was all of the manifests?

The effect of this PR is that that `lint_from_helm` lints nothing?

I'm not sure what the intention is behind this PR?

JamesDoingStuff: I think I might have confused things when I created the ticket for this without having looked through the CLI code. I read `/tmp` and assumed it was linting something that wasn't meant to be included - I didn't realise that that was how helm-based workflows are handled.
Rather than filtering the `tmp` workflows out, I think the purpose of this ticket should instead be to make sure that the name of the helm workflow gets passed to the report - `Template Name: /tmp/argo-lint/workflow_2.yaml - Errors found` isn't very helpful

hazdl: As I understand it, the linting process checks the temporary folder before any files are generated, which results in a linting error because the folder is empty. Since the folder is populated at runtime, I excluded it from the linting checks.

JamesDoingStuff: If it does fail with some kind of 'failed to find any workflows' error then I think that would be a separate issue. Contrary to what I wrote in the ticket, we do want these tmp files to be linted - they correspond to the [helm-based workflow templates](https://github.com/DiamondLightSource/workflows/tree/main/examples/helm-based-templates/templates). These workflows can't be linted with the standard Argo Workflow linter because of the helm templating, so they must be parsed using `helm template` first - this creates the tmp files at runtime, and these can be linted normally

PR #1380, opened 2026-06-09
drh/fix identity mapper

davehadley:


.github/workflows/_identity_mapper_code.yaml
```diff
@@ -0,0 +1,37 @@
...
+    name: python
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@08807647e7069bb48b6ef5acd8ec9567f424441b # v8.1.0
+        with:
+          version: "0.11.19"
```
JamesDoingStuff, line 17: If we use `working-directory: backend/identity-mapper` could we have this action get the uv version from the `pyproject.toml` instead? Might be easier to keep the two in line if so, even if it's just one fewer dependabot PR to click on

backend/identity-mapper/src/identity_mapper/_sync_ldap_to_kubernetes.py
```diff
@@ -31,7 +30,8 @@ def sync_ldap_to_kubernetes(
         return
 
     _logger.info(
-        "Applying %d changes (%d creates, %d deletes, %d updates) to match desired LDAP state...",
+        "Applying %d changes (%d creates, %d deletes, %d updates)"
+        "to match desired LDAP state...",
```
JamesDoingStuff:
```suggestion
        "Applying %d changes (%d creates, %d deletes, %d updates) "
        "to match desired LDAP state...",
```
if this concats how I think it will. Sorry for nitpicking!

JamesDoingStuff (approved):

.github/workflows/_identity_mapper_code.yaml
davehadley: This is the version of `uv` CLI that is used to run CI lints and tests. This follows the recommendations at: <https://docs.astral.sh/uv/guides/integration/github/>

I don't think that the uv version is specified in the pyproject.toml. Only `uv_build`.

backend/identity-mapper/src/identity_mapper/_sync_ldap_to_kubernetes.py
davehadley: fixed

PR #1386, opened 2026-06-11
feat(events): allow stomp server to handle multiple event sources

JamesDoingStuff:
[AP-1158](https://jira.diamond.ac.uk/browse/AP-1158)
Each new connection is stored in a connection manager object, each with its own event queue. This means that a new connection won't overwrite a previous one

### DIFF
--- a/charts/events/generic-servers/stomp-server.py
+++ b/charts/events/generic-servers/stomp-server.py
@@ -24,8 +24,8 @@ class EventSourceConfig(BaseModel):
 
 def parse_config(config: bytes) -> EventSourceConfig | None:
     try:
-        decoded_confid = config.decode()
-        config_entries = decoded_confid.strip().split("\n")
+        decoded_config = config.decode()
+        config_entries = decoded_config.strip().split("\n")
         config_dict = dict(entry.split(": ") for entry in config_entries)
         return EventSourceConfig.model_validate(config_dict)
     except:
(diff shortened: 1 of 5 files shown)


charts/events/generic-servers/stomp-server.py
```diff
@@ -37,6 +37,38 @@ def _build_event(payload: str) -> generic_pb2.Event:
     return generic_pb2.Event(name="workflow-trigger", payload=payload.encode())
 
 
+class ConnectionManager():
+    def __init__(self):
+        self.connection: stomp.Connection
+        self.queue = queue.SimpleQueue()
+
+    def create_connection(self, request: generic_pb2.EventSource):
```
davehadley: suggest to type hint return types.

davehadley (approved):

charts/events/templates/workflow-template.yaml
TBThomas56, line 1: Is there a reason why the ClusterWorkflowTemplate is in the templates in events and not in examples? Is this tracked by groups or is it intentionally ignored?

charts/events/templates/example-trigger.yaml
```diff
@@ -33,6 +33,23 @@ metadata:
...
+---
+apiVersion: workflows.diamond.ac.uk/v1alpha1
+kind: Trigger
+metadata:
+  name: visr-trigger
+  labels:
+    workflows.diamond.ac.uk/source: generic
+    workflows.diamond.ac.uk/beamline: b01-1
+spec:
+  enabled: true
```
TBThomas56, line 51: would it be worth helm templating the different trigger.enabled values?

JamesDoingStuff: Yeah I think I will put that in as it will mean I can test events more easily without spamming dozens of workflows. Worth noting though that the plan is to eventually have these Triggers be created by users via the graph, so these will get deleted

charts/events/templates/workflow-template.yaml
JamesDoingStuff: These are purely for testing event-based triggering while it's still in development and aren't meant to be examples for users to follow, so I don't intend to keep them in the repository long-term. They aren't tracked by `groups` but they aren't ignored either; they still appear in the template list.

charts/events/templates/example-trigger.yaml
JamesDoingStuff: Actually I've just realised that the `enabled` field in the triggers currently doesn't do anything - I need to implement it in the metacontroller sync hook. I'll probably do that in a separate PR as it's a little beyond the scope of this one I feel

PR #1388, opened 2026-06-12
fix(charts): remove erroneous label on sealed secret causing Argo CD sync issues

davehadley:


backend/Cargo.lock
```diff
@@ -2733,7 +2733,7 @@ dependencies = [
 
 [[package]]
 name = "graph-proxy"
-version = "0.1.18"
+version = "0.1.19"
```
JamesDoingStuff, line 2736: I don't know if this was intentionally included but I guess it needs to be done at some point since it didn't happen in #1263

JamesDoingStuff (approved):

backend/Cargo.lock
davehadley: thanks. this was an accident, I will split it out so it goes in it's own commit

PR #1390, opened 2026-06-12
chore(charts): remove file accidentally included in 177a467

davehadley:


TBThomas56 (changes requested):
Since there are secrets in this commit we do not want, I think it might be better to completely remove it from the history using `git filter-repo --path <path/to/file> --invert-paths`

Alternatively, can these secrets be changed?

JamesDoingStuff (approved):

TBThomas56 (approved):

